PR #9, opened 2017-03-28
Add file stream

ericxtang:
Adding "FileStream" so a video stream can be written to disk.  This was used to test for SRS transcoding consistency.  The result looks good so far.


ericxtang:
Yeah I just ran the streaming by hand and inspected the output by checking the md5 of all the segments

PR #11, opened 2017-04-07
Video segment

ericxtang:
The important changes here are the segmenter and the listener.

Segmenter uses FFMpeg to cut video into HLS segments.

Also, the listener automatically uses the segmenter to create a HLS stream whenever a RTMP stream is published.


segmenter/video_segmenter_test.go
```diff
@@ -0,0 +1,202 @@
...
+		dst.WritePacket(pkt)
+	}
+}
+func (s *TestStream) WriteRTMPToStream(ctx context.Context, src av.DemuxCloser) error { return nil }
+func (s *TestStream) WriteHLSPlaylistToStream(pl m3u8.MediaPlaylist) error            { return nil }
+func (s *TestStream) WriteHLSSegmentToStream(seg stream.HLSSegment) error             { return nil }
+func (s *TestStream) ReadHLSFromStream(buffer stream.HLSMuxer) error                  { return nil }
+
+func TestSegmenter(t *testing.T) {
+	workDir := "/Users/erictang/Code/go/src/github.com/livepeer/lpms/segmenter/tmp" // "/Users/erictang/Code/go/src/github.com/livepeer/go-livepeer/livepeer/storage/streaming/tmp"
```
dob: Should we make this a directory not specific to your machine so that the tests can run on CI and elsewhere?

vidlistener/listener.go
```diff
@@ -17,38 +22,90 @@ type VidListener struct {
...
+					return err
+				}
+				ss := stream.HLSSegment{Data: seg.Data, Name: seg.Name}
+				glog.Infof("Writing stream: %v, len:%v", ss.Name, len(seg.Data))
+				hs.WriteHLSSegmentToStream(ss)
+			}
+		}()
+	}()
+	select {
+	case err := <-c:
```
dob: Can this listener be killed in any way, or does it just block forever? Or does PollSegment always eventually return an error?

cmd/rtmp_segment_packer/main.go
```diff
@@ -0,0 +1,344 @@
...
+			time.Sleep(time.Millisecond * 5)
+
+			if err != nil {
+				glog.Infof("Error writing RTMP packet from Stream %v to mux", s.StreamID)
+				dst.WriteTrailer()
+				return err
+			}
+		}
+
+		glog.Infof("Writing seg 6")
```
ericxtang: This is just a test.  I was trying to examine individual segments and see if transcoding output would be consistent if you skipped certain segments - and turns out no.  This made me pursue an alternative approach where we use FFMpeg to segment the video first.

lpms.go
```diff
@@ -65,10 +65,10 @@ func (l *LPMS) Start() error {
 //HandleRTMPPublish offload to the video listener
 func (l *LPMS) HandleRTMPPublish(
 	getStreamID func(reqPath string) (string, error),
-	stream func(reqPath string) (stream.Stream, error),
+	getStream func(reqPath string) (stream.Stream, stream.Stream, error),
```
ericxtang: getStream returns 2 streams, one for RTMP, and one for HLS (which the segmenter will populate).  This code change means the HLS stream will always be ready when a RTMP stream is published to the node. (But it won't be sent through the network unless requested.)

segmenter/video_segmenter.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+func pollPlaylist(ctx context.Context, fn string, sleepTime time.Duration, lastFile []byte) (f []byte, err error) {
+	for {
+		if _, err := os.Stat(fn); err == nil {
+			if err != nil {
+				return nil, err
+			}
+
+			content, err := ioutil.ReadFile(fn)
```
ericxtang: Not right now... I agree it can be a problem - was thinking about adding a "clear history" button in the UI or something...

ericxtang:
I'm confused about this new UI... It says there is a change request, but I've already made the changes (about fixing the tests...)  Is it working for you now?

dob:
:ship:

PR #13, opened 2017-04-14
Video segment

ericxtang:
Most of this PR is the StreamSubscriber, which allows a subscription workflow for any VideoStream, so a stream can be broadcasted to multiple subscribers.

### DIFF
--- a/stream/hls.go
+++ b/stream/hls.go
@@ -13,7 +13,7 @@ var ErrNotFound = errors.New("Not Found")
 
 type HLSDemuxer interface {
 	//This method should ONLY push a playlist onto a chan when it's a NEW playlist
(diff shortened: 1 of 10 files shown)


stream/hls.go
```diff
@@ -58,11 +58,15 @@ func (b *HLSBuffer) WriteSegment(name string, s []byte) error {
...
-
+		b.lock.Lock()
 		if b.plCacheNew {
-			return b.plCache, nil
+			defer b.lock.Unlock()
+
 			b.plCacheNew = false
+			return b.plCache, nil
 		}
+		b.lock.Unlock()
```
dob: Why not just defer this immediately after locking? Because the 1 second sleep can cause a deadlock?

PR #15, opened 2017-04-19
desktop app integration

ericxtang:
Add the video format when creating the video stream - need this info for the app (also will be useful for the node)


ericxtang:
@dob could be - but RTMP / HLS consumption models are very different (one is a single stream, the other keeps requesting updates).  I tried to have a unified stream concept at the beginning but the code got complicated when you are trying to consume the stream.  But this could change as we are adding more video formats.

PR #17, opened 2017-04-25
add ffmpegPath, bugfixes

ericxtang:
segment length of 2 was causing a lot of buffering.  Increasing to 5.

### DIFF
--- a/segmenter/video_segmenter.go
+++ b/segmenter/video_segmenter.go
@@ -11,6 +11,8 @@ import (
 	"strconv"
 	"time"
 
+	"path"
+
 	"github.com/golang/glog"
 	"github.com/kz26/m3u8"
 	"github.com/livepeer/lpms/stream"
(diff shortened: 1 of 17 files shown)


segmenter/video_segmenter.go
```diff
@@ -82,7 +83,15 @@ func (s *FFMpegVideoSegmenter) RTMPToHLS(ctx context.Context, opt SegmenterOptio
...
 
 	//This command needs to be manually killed, because ffmpeg doesn't seem to quit after getting a rtmp EOF
-	cmd := exec.Command("ffmpeg", "-i", s.LocalRtmpUrl, "-vcodec", "copy", "-acodec", "copy", "-bsf:v", "h264_mp4toannexb", "-f", "segment", "-muxdelay", "0", "-segment_list", plfn, tsfn)
+	glog.Infof("Ffmpeg path: %v", s.ffmpegPath)
+
+	var cmd *exec.Cmd
+	if s.ffmpegPath == "" {
+		cmd = exec.Command("ffmpeg", "-i", s.LocalRtmpUrl, "-vcodec", "copy", "-acodec", "copy", "-bsf:v", "h264_mp4toannexb", "-f", "segment", "-muxdelay", "0", "-segment_list", plfn, tsfn)
+	} else {
+		cmd = exec.Command(s.ffmpegPath+"/ffmpeg", "-i", s.LocalRtmpUrl, "-vcodec", "copy", "-acodec", "copy", "-bsf:v", "h264_mp4toannexb", "-f", "segment", "-muxdelay", "0", "-segment_list", plfn, tsfn)
```
dob: should probably use `path.Join()` for cross platform compatibility and for handling the extra/missing `/` case

vidplayer/player.go
```diff
@@ -63,28 +65,34 @@ func handleHLS(w http.ResponseWriter, r *http.Request, getHLSBuffer func(reqPath
...
+		//Remove all but the last 5 segments
 		c := 0
 		for _, seg := range pl.Segments {
 			if seg != nil {
 				// segs = append(segs, seg)
 				c = c + 1
 			}
 		}
-		for c > 2 {
+		for c > 5 {
```
dob: Segment length and the player buffer length are probably good candidates for constant variables instead of hardcoded here since we probably will want to reuse them throughout LPMS.

dob:
🚢

PR #20, opened 2017-05-01
HLS changes

ericxtang:
This change is related to how HLS is treated in LPMS.

* We add a buffer cap limit, so it doesn't just keep adding HLS segments
* HLSBuffers can now generate playlists from its segments
* Add methods to synchronously read HLS from a VideoStream

### DIFF
--- a/stream/hls.go
+++ b/stream/hls.go
@@ -3,19 +3,25 @@ package stream
 import (
 	"context"
 	"errors"
+	"sort"
+	"strconv"
+	"strings"
 	"sync"
 	"time"
 
+	"github.com/golang/glog"
 	"github.com/kz26/m3u8"
 )
 
 var ErrNotFound = errors.New("Not Found")
+var ErrBadHLSBuffer = errors.New("BadHLSBuffer")
 
 type HLSDemuxer interface {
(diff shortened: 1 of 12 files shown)


stream/hls.go
```diff
@@ -25,18 +31,19 @@ type HLSMuxer interface {
...
-	plCache m3u8.MediaPlaylist
-	sq      *ConcurrentMap
-	lock    sync.Locker
+	plCache  m3u8.MediaPlaylist
+	sq       *ConcurrentMap
+	lock     sync.Locker
+	Capacity uint
 }
 
-func NewHLSBuffer() *HLSBuffer {
```
dob: Should we just leave this method and invoke the new method with a default capacity?

ericxtang: Golang doesn't have method overload, which makes things a little annoying to maintain (cannot just add a new param in the constructor without changing existing code).  Not sure what's the best practice here.  I can add a "NewHLSBufferWithCap", but maybe it's good to enforce the cap for this case (following the same pattern as grafov/m3u8).

ericxtang:
Also removed lpms binary

PR #30, opened 2017-07-03
Interface cleanup / transcoder

ericxtang:
A lot of changes here - 
* A new FFmpegSegmentTranscoder, invokes ffmpeg for transcoding individual segments.
* Interface cleanup - RTMP listener no longer automatically segments the stream into HLS.


ericxtang:
There is an issue for doing the native ffmpeg integration.

I need to fix the tests on circleci.  We also have some race conditions in the test, so they don't always pass.  Will make an issue for this.

PR #52, opened 2018-02-09
Transcoding via FFmpeg API

j0sh:
Commits roughly broken down by feature for easier review.

### DIFF
--- a/ffmpeg/ffmpeg.go
+++ b/ffmpeg/ffmpeg.go
@@ -9,7 +9,7 @@ import (
 	"unsafe"
 )
 
-// #cgo pkg-config: libavformat
+// #cgo pkg-config: libavformat libavfilter
 // #include <stdlib.h>
 // #include "lpms_ffmpeg.h"
 import "C"
--- a/ffmpeg/ffmpeg.go
+++ b/ffmpeg/ffmpeg.go
@@ -1,7 +1,11 @@
 package ffmpeg
 
 import (
+	"errors"
+	"fmt"
 	"github.com/golang/glog"
+	"strconv"
+	"strings"
 	"unsafe"
 )
 
@@ -9,6 +13,8 @@ import (
 // #include <stdlib.h>
 // #include "lpms_ffmpeg.h"
 import "C"
+
+var ErrTranscoderRes = errors.New("TranscoderInvalidResolution")
 
 func RTMPToHLS(localRTMPUrl string, outM3U8 string, tmpl string, seglen_secs string) error {
 	inp := C.CString(localRTMPUrl)
@@ -22,6 +28,41 @@ func RTMPToHLS(localRTMPUrl string, outM3U8 string, tmpl string, seglen_secs str
 	C.free(unsafe.Pointer(seglen))
 	if 0 != ret {
 		glog.Infof("RTMP2HLS Transmux Return : %v\n", Strerror(ret))
+		return ErrorMap[ret]
+	}
+	return nil
+}
+
+func Transcode(input string, ps []VideoProfile) error {
+	inp := C.CString(input)
+	params := make([]C.output_params, len(ps))
+	for i, param := range ps {
+		oname := C.CString(fmt.Sprintf("out%v%v", i, input))
+		res := strings.Split(param.Resolution, "x")
+		if len(res) < 2 {
+			return ErrTranscoderRes
+		}
+		w, err := strconv.Atoi(res[0])
+		if err != nil {
+			return err
+		}
+		h, err := strconv.Atoi(res[1])
+		if err != nil {
+			return err
+		}
+		br := strings.Replace(param.Bitrate, "k", "000", 1)
+		bitrate, err := strconv.Atoi(br)
+		if err != nil {
+			return err
+		}
+		fps := C.AVRational{num: C.int(param.Framerate), den: 1}
+		params[i] = C.output_params{fname: oname, fps: fps,
+			w: C.int(w), h: C.int(h), bitrate: C.int(bitrate)}
+	}
+	ret := int(C.lpms_transcode(inp, (*C.output_params)(&params[0]), C.int(len(params))))
+	C.free(unsafe.Pointer(inp))
+	if 0 != ret {
+		glog.Infof("Transcoder Return : %v\n", Strerror(ret))
 		return ErrorMap[ret]
 	}
 	return nil
--- a/ffmpeg/ffmpeg.go
+++ b/ffmpeg/ffmpeg.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"github.com/golang/glog"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"unsafe"
@@ -37,7 +38,7 @@ func Transcode(input string, ps []VideoProfile) error {
 	inp := C.CString(input)
 	params := make([]C.output_params, len(ps))
 	for i, param := range ps {
-		oname := C.CString(fmt.Sprintf("out%v%v", i, input))
+		oname := C.CString(fmt.Sprintf("out%v%v", i, filepath.Base(input)))
 		res := strings.Split(param.Resolution, "x")
 		if len(res) < 2 {
 			return ErrTranscoderRes
--- a/ffmpeg/lpms_ffmpeg.c
+++ b/ffmpeg/lpms_ffmpeg.c
@@ -1,12 +1,15 @@
 #include "lpms_ffmpeg.h"
 
 #include <libavformat/avformat.h>
+#include <libavutil/opt.h>
 
 //
 // Internal transcoder data structures
 //
 struct input_ctx {
   AVFormatContext *ic; // demuxer required
+  AVCodecContext  *vc; // video decoder optional
+  int vi; // video stream index
 };
 
 struct output_ctx {
@@ -171,6 +174,7 @@ open_output_err:
 static void free_input(struct input_ctx *inctx)
 {
   if (inctx->ic) avformat_close_input(&inctx->ic);
+  if (inctx->vc) avcodec_free_context(&inctx->vc);
 }
 
 static int open_input(char *inp, struct input_ctx *ctx)
@@ -180,6 +184,7 @@ static int open_input(char *inp, struct input_ctx *ctx)
   fprintf(stderr, msg); \
   goto open_input_err; \
 }
+  AVCodec *codec = NULL;
   AVFormatContext *ic   = NULL;
 
   // open demuxer
@@ -189,6 +194,20 @@ static int open_input(char *inp, struct input_ctx *ctx)
   ret = avformat_find_stream_info(ic, NULL);
   if (ret < 0) dd_err("Unable to find input info\n");
 
+  // open video decoder
+  ctx->vi = av_find_best_stream(ic, AVMEDIA_TYPE_VIDEO, -1, -1, &codec, 0);
+  if (ctx->vi < 0) {
+    fprintf(stderr, "No video stream found in input\n");
+  } else {
+    AVCodecContext *vc = avcodec_alloc_context3(codec);
+    if (!vc) dd_err("Unable to alloc video codec\n");
+    ctx->vc = vc;
+    ret = avcodec_parameters_to_context(vc, ic->streams[ctx->vi]->codecpar);
+    if (ret < 0) dd_err("Unable to assign video params\n");
+    ret = avcodec_open2(vc, codec, NULL);
+    if (ret < 0) dd_err("Unable to open video decoder\n");
+  }
+
   return 0;
 
 open_input_err:
@@ -197,7 +216,7 @@ open_input_err:
 #undef dd_err
 }
 
-int process_in(struct input_ctx *ictx, AVPacket *pkt)
+int process_in(struct input_ctx *ictx, AVFrame *frame, AVPacket *pkt)
 {
 #define dec_err(msg) { \
   if (!ret) ret = -1; \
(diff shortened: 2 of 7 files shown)


j0sh:
Ready for review.

ffmpeg/ffmpeg.go
```diff
@@ -27,6 +33,37 @@ func RTMPToHLS(localRTMPUrl string, outM3U8 string, tmpl string, seglen_secs str
 	return nil
 }
 
+func Transcode(input string, ps []VideoProfile) error {
+	inp := C.CString(input)
+	params := make([]C.output_params, len(ps))
+	for i, param := range ps {
```
ericxtang: I think we have to put the transcode call in a go routine somewhere in this call stack.  Otherwise it'll block until the work finishes.  Maybe the best place is in `Livepeernode.TranscodeAndBroadcast()` - just pointing it out so we don't forget.

j0sh: Sure -- I believe the old behavior was synchronous as well. If the blocking has been a problem with with go-livepeer all along, then we can change it there.

ffmpeg/lpms_ffmpeg.c
```diff
@@ -102,3 +140,588 @@ int lpms_rtmp2hls(char *listen, char *outf, char *ts_tmpl, char* seg_time)
...
+  if (ictx->vc) {
+    codec = avcodec_find_encoder_by_name("libx264"); // XXX make more flexible?
+    if (!codec) em_err("Unable to find libx264");
+
+    // open video encoder
+    // XXX use avoptions rather than manual enumeration
+    vc = avcodec_alloc_context3(codec);
+    if (!vc) em_err("Unable to alloc video encoder\n"); // XXX shld be optional
+    octx->vc = vc;
+    vc->width = av_buffersink_get_w(octx->vf.sink_ctx);
```
ericxtang: Anywhere I can learn about the buffer source / buffer sink API?

ffmpeg/lpms_ffmpeg.c
```diff
@@ -102,3 +140,588 @@ int lpms_rtmp2hls(char *listen, char *outf, char *ts_tmpl, char* seg_time)
...
+      struct filter_ctx *filter = NULL;
+      AVStream *ost = NULL;
+      AVCodecContext *encoder = NULL;
+
+      if (ist->index == ictx.vi && ictx.vc) {
+        ost = octx->oc->streams[0];
+        encoder = octx->vc;
+        filter = &octx->vf;
+      } else if (ist->index == ictx.ai && ictx.ai) {
+        ost = octx->oc->streams[!!ictx.vc];
```
ericxtang: Why !!ictx.vc here instead of something like `octx->oc->streams[1]`?

ffmpeg/lpms_ffmpeg.c
```diff
@@ -102,3 +140,588 @@ int lpms_rtmp2hls(char *listen, char *outf, char *ts_tmpl, char* seg_time)
...
+  if (ret < 0) proc_err("Error writing frame\n"); // XXX handle better?
+
+proc_cleanup:
+  av_frame_unref(frame);
+  av_packet_unref(&pkt);
+  return ret;
+#undef proc_err
+}
+
+#define MAX_OUTPUT_SIZE 3
```
ericxtang: Why use MAX_OUTPUT_SIZE if we already have nb_outputs?

j0sh: Good catch -- we preallocate MAX_OUTPUT_SIZE output contexts (on the stack), but we're missing the check for nb_outputs <= MAX_OUTPUT_SIZE.

ffmpeg/lpms_ffmpeg.c
j0sh: Accommodates the case where we don't have video. Video streams are always at index 0, but audio may be 0 or 1 depending on whether `ictx.vc` is non-null.

ffmpeg/lpms_ffmpeg.c
j0sh: Probably the canonical source for those specific functions is the header documentation, the comments are fairly thorough: https://github.com/FFmpeg/FFmpeg/blob/master/libavfilter/buffersink.h

j0sh:
Happy to explain whatever else about the code that may need clarification.

Agreed that the filtering API is a bit unexplored by the community, partially due to its relative new-ness. It's also a bit cumbersome to use compared to straight swscale and swresample, but I think the flexibility offered by filters will pay off in the long run. Filters have already been useful to avoid re-implementing fps handling, as well as certain adjustments on the audio (but the extra audio filters are disabled in the current code, aside from resampling).

For me, the best source of information has been the lavfi header documentation itself [1] , plus the docs on specific filters themselves [2] and the sample code [3][4].

[1] https://github.com/FFmpeg/FFmpeg/blob/master/libavfilter/avfilter.h
[2] http://ffmpeg.org/ffmpeg-filters.html
[3] https://github.com/FFmpeg/FFmpeg/blob/master/doc/examples/filtering_video.c
[4] https://github.com/FFmpeg/FFmpeg/blob/master/doc/examples/filter_audio.c

j0sh:
Fixed / clarified a couple of the points that were mentioned, and added tests for those cases to boot. Which is a good thing, because the tests uncovered bugs (which have also been fixed). There is also a small memory leak sometimes (about one frame's worth of memory) but will track that down.

ffmpeg/lpms_ffmpeg.c
j0sh: Added a comment explaining as much, and test cases to ensure that it actually handles no-audio and no-video correctly. Note that this line contains a bug:

`} else if (ist->index == ictx.ai && ictx.ai) {`

which should be `} else if (ist->index == ictx.ai && ictx.ac) {`

ericxtang:
🚢  after addressing a small issue - **I ran test.sh in devenv and kept getting a left-over ts segment in `/transcoder`.  Seems like it's not being properly cleaned up**

This has been a very educational code review process for me.  I still can't say I'd be able to build this from scratch, but at least now I understand some of the basic concepts in libav.

This note was helpful for me as I was going through the codebase - 

open_output

* av_guess_format
* avformat_alloc_output_context2
* If we have input video context
    * get the x264 codec (which is also the encoder)
    * vc = avcodec_alloc_context3 based on the codec
    * set vc's parameters based on octx (width, height, framerate, time_base, thread count, pix_fmt)
        * some of these are retrieved by calling av_buffersink_get_x.  
    * avcodec_open2 (what is this for? why do you have to “open” the codec?)
    * avformat_new_stream(call on the output context)
    * avcodec_parameters_from_context
* If we have input audio context
    * get the acc codec (which is also the encoder)
    * ac = avcodec_alloc_context3 - open the codec
    * set ac's parameters based on octx
        * a lot of these are retrieved by calling av_buffersink_get_x.
    * avcodec_open2
    * avformat_new_stream
    * avcodec_parameters_from_context
    * set timebase
* avio_open
* av_format_write_header

open_input (initialize everything in input_ctx)

* avformat_open_input - create ic (input AVFormatContext)
* avformat_find_stream_info
* open video decoder
    * av_find_best_stream (get the video stream)
    * vc = avcodec_alloc_context3 (AVCodecContext)
    * avcodec_parameters_to_context(vc, ic→streams[ctx→vi]→codecpar
    * av_opt_set_int (what does this do?)
    * avcodec_open2
* open audio decoder
    * av_find_best_stream (get the audio stream)
    * ac = avcodec_alloc_context3
    * avcodec_parameters_to_context
    * avcodec_open2

init_video_filters

* allocates a filter graph (avfilter_graph_alloc)
* create the “source” of the filter by adding a buffer to the graph, passing string parameters to it
* create the “sink” of the filter by adding a buffer sink to the graph, where we will get our filtered frames from
* use the magical function call “avfilter_graph_parse_ptr” to create the graph with the string-based filter graph, a filter input and a filter output
* use avfilter_graph_config to make sure the graph is valid
* allocate the filter output frame with av_frame_alloc

init_audio_filters

* Pretty much works the same way as init_video_filters

process_in (put everything into the frame?)

* av_init_packet
* loop
    * av_read_frame(ictx→ic, pkt) - read from input context into the packet
    * pkt has a stream_index → it can be either a video packet or an audio packet
    * get the decoder depending on the pkt type (either video or audio decoder)
    * avcodec_send_packet(put the packet into the decoder)
    * avcodec_receive_frame(put the decoder content into the frame)
* at the end (AVERROR_EOF), flush the decoder
    * call avcodec_send_packet with NULL
    * call avcodec_receive_frame
    * why set pkt→stream_index?
    * why not call dec_cleanup after the flush?

process_out

* depending on if the filter is active, we do “filtering” on the input frame
    * av_frame_copy_props(frame, inf) → copy the inf into frame
    * av_buffersrc_add_frame_flags (what does this do?)
    * av_buffersink_get_frame(get the frame out of the sink_ctx?)
    * tb = av_buffersink_get_frame_time_base(filter→sink_ctx)
* av_init_packet
* if the encoder is set (when should we NOT set the encoder?)
    * avcodec_send_frame
    * avcodec_receive_packet
* set pkt.pts, pkt.dts, pkt.duration
* av_interleaved_write_frame (from packet into octx->oc)

lpms_transcode

* open_input with ictx (initialize every thing in ictx)
* memset outputs
* for each output profile
    * sets context params (fname, width, height, bitrate, fps)
    * init_video_filters
    * init_audio_filters
    * open_output (initialize everything in octx)
* av_init_packet
* av_frame_alloc
* while loop
    * process_in → demux, decode (put everything into the frame)
    * get input stream (could be video or audio)
    * for each output profile
        * get output stream, encoder, and filter (could be video or audio, depending on input stream)
        * process_out
* flush
    * process_out with NULL (instead of frame)
    * av_interleaved_write_frame on oc
    * av_write_trialer on oc

j0sh:
Thanks for the review and the questions, helps to re-visit some of the assumptions I may have been making when writing this.

> I ran test.sh in devenv and kept getting a left-over ts segment in /transcoder. Seems like it's not being properly cleaned up

Good catch, this was actually a latent issue in the old code as well -- if ffmpeg failed to start for any reason then it would have failed to clean up. Fixed.

@ericxtang  what do you think about changing this further to stop creating temporary files in `ffmpeg_segment_transcoder` and just take an existing filename? This would require minor changes to go-livepeer as well.

> avcodec_open2 (what is this for? why do you have to “open” the codec?)

There are two parts here: the AVCodec and the AVCodecContext.

* The AVCodec identifies the specific codec and its constraints (codec name, media type, parameters that the en/decoder supports such as the selection of pixel formats or sample rates, etc).

* The AVCodecContext is the actual instantiation of the de/encoder that is operated on. So we first find the AVCodec, then instantiate the AVCodecContext based on the parameters we need (and what the codec supports).

> av_opt_set_int (what does this do?)

This sets the decoded frames to be reference counted rather than following the old semantics of copying everything if some data needs to be retained outside the decode process. Actually it's deprecated right now due to our usage of the receive_frame API; it was cargo-culted from some libavfilter examples. Removed.

We actually should be setting all our optional fields (bitrate, sample rate, etc etc) via AVOption, added some more info here https://github.com/livepeer/lpms/issues/55#issuecomment-369030155

> init_audio_filters Pretty much works the same way as init_video_filters

Eventually, we should should look into refactoring this so there's less boilerplate (avfilter_graph_parse2 looks promising) and more shared code between the audio and video filters.

> why set pkt→stream_index?

When we flush, it's because we've EOF'd on the input. However, we may still have some frames pending, so setting pkt->stream_index during flushing is how we identify which (input) stream the flushed frame belongs to, so we can select the correct encoders for the output after the process_in function returns.

Normally doing it this way is safer than poking inside AVFrame itself to determine the content type (whether audio/video), because decoded AVFrames don't carry any information about the source packet/stream. This is problematic for eg, multiple audio or video streams in the same output container.

> why not call dec_cleanup after the flush?

dec_cleanup releases the reference to the AVPacket that was read during input, but when flushing we don't have any more AVPackets to read. Rather, we're trying to extract the last remaining frames that may be buffered within the decoder.

> av_buffersrc_add_frame_flags (what does this do?)

This is a rather confusingly named function, but it should be read like av_buffersrc_add_frame_{with}_flags . All it does is feed the buffersrc with the frame and some additional flags. Changed it to use a helper function that's a bit less explicit, but follows the same semantics: `av_buffersrc_write_frame` .

> av_buffersink_get_frame(get the frame out of the sink_ctx?)

Yep, gets the frame out of the filter graph (the buffersink being the 'exit point' for that).

> if the encoder is set (when should we NOT set the encoder?)

Right now it should always be set unless the file doesn't actually have an audio or video stream. That check might have been a bit of leftover debugging.

In general we can also skip encoding if we're just doing stream copy / transmuxing , although lpms_transcode doesn't support that right now (and when it does, we might want to do it in a separate code path for clarity). Once stream copy is in, we can also subsume lpms_rtmp2hls into the lpms_transcode and save another ~100 lines or so.

ericxtang:
Thanks for the answers @j0sh.  I'm not sure what you mean by "just take an existing filename" - but I remember chatting about removing the temporary file approach with you and @dob but we decided it's still useful to have the temporary files.  Maybe we can use some kind of naming convention to organize the video files so we can keep them around in case the user wants to save the video.  What do you think?

j0sh:
About the temporary files -- we'd still have them, but those would be written to disk around here https://github.com/livepeer/go-livepeer/blob/b54cc54975782608266ad44fdbbe509df261c58e/core/livepeernode.go#L240

rather than within the LPMS transcode function.

The LPMS transcoder signature could change from https://github.com/livepeer/lpms/blob/master/transcoder/ffmpeg_segment_transcoder.go#L27 to something like:

`func (t *FFMpegSegmentTranscoder) Transcode(fname string) ([][]byte, error)`

We can keep the return value as bytes for now, since I think changing that might require more thought.

The net effect would be to prevent writing additional test segments to disk, but it's a very minor issue since we don't yet have that many tests around transcoding.

j0sh:
Changed the API here https://github.com/livepeer/lpms/pull/52/commits/61305278a1d19e3028e02fbded6fb72c283b6cf9 and added tests here https://github.com/livepeer/lpms/pull/52/commits/444fb87005db5e9965a1f5da317d6868c0fb3e59

ericxtang:
:shipit: :shipit: :shipit:

PR #56, opened 2018-03-01
Fix breakage exposed by goclient.

j0sh:
Should have added internal LPMS tests for these cases as well, but getting this PR out for the sake of expediency so we can merge this into the goclient.

### DIFF
--- a/cmd/example/main.go
+++ b/cmd/example/main.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"io/ioutil"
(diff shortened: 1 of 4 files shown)


ericxtang:
🚢

j0sh:
Oy, you're right. Was focused on the test cases and didn't think to check the example program. Fixed compilation, but haven't actually tried to run it.

https://github.com/livepeer/lpms/pull/56/commits/b0845184da7dd7094b3f958065fd4b847ddc6ae2

PR #59, opened 2018-03-06
segmenter: Drop video frames until the first keyframe.

j0sh:


ffmpeg/lpms_ffmpeg.c
```diff
@@ -114,6 +116,13 @@ int lpms_rtmp2hls(char *listen, char *outf, char *ts_tmpl, char* seg_time)
     ist = ic->streams[pkt.stream_index];
     ost = oc->streams[stream_map[pkt.stream_index]];
     int64_t dts_next = pkt.dts, dts_prev = prev_ts[pkt.stream_index];
+    if (oc->streams[pkt.stream_index]->codecpar->codec_type == AVMEDIA_TYPE_VIDEO) {
+        if (AV_NOPTS_VALUE == dts_prev) {
+          if (pkt.flags & AV_PKT_FLAG_KEY) got_video_kf = 1;
+          else goto r2hloop_end; // XXX accommodate audio-only streams
```
yondonfu: Is this additional goto needed? Seems like if we do not find a video keyframe in the check on L121 execution would continue to the check on L125 and then evaluate the goto there to jump to `r2hloop_end`?

j0sh: Good catch, fixed.

segmenter/video_segmenter_test.go
```diff
@@ -405,3 +408,58 @@ func TestSegmentDefaults(t *testing.T) {
...
+	// actually segment
+	ffmpeg.InitFFmpeg()
+	defer ffmpeg.DeinitFFmpeg()
+	err = ffmpeg.RTMPToHLS(fname, oname, path.Join(dir, "out")+"_%d.ts", "4")
+	if err != nil {
+		t.Errorf("Error segmenting - %v", err)
+		return
+	}
+	// and now check that segmented result does have keyframe at beginning
+	out, err = ffprobe_firstframeflags(path.Join(dir, "out_0.ts"))
```
yondonfu: Just for personal clarification...is the idea that if the input file does not have a video keyframe, the segmenter will make sure that the first segment does have a keyframe?

j0sh: Correct, the segmenter will drop the packets until it gets a keyframe. Truth be told, this isn't a great solution for a few reasons:

* If we don't have video it drops everything. (The same has been true for audio though.)
* It drops audio that may come in before video, or "late" video
* If it doesn't get a video keyframe, it drops everything also. This video is probably un-decodeable anyway, but it may still be useful to write just the audio to disk.

Fixing these the correct way will take a bit more work than this short-term solution.

Related: https://github.com/livepeer/lpms/issues/60

yondonfu: Ah ok got it yeah I confused myself previously - I was thinking that the segmenter somehow adds a keyframe when in reality it just drops packets until it gets a keyframe such that the result is that the first segment outputted by the segmenter necessarily contains a keyframe

j0sh:
Worked around the test so that it doesn't require a new ffmpeg build anymore. A bit more brittle than the earlier method but we're green for now :+1:

PR #65, opened 2018-03-31
Segment length checker.

j0sh:
For https://github.com/livepeer/go-livepeer/issues/344

### DIFF
--- a/ffmpeg/ffmpeg.go
+++ b/ffmpeg/ffmpeg.go
@@ -70,6 +70,23 @@ func Transcode(input string, workDir string, ps []VideoProfile) error {
 	return nil
 }
 
+// Check media length up to given limits for timestamp (ms) and packet count.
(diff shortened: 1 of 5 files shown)


ericxtang:
How do you envision `CheckMediaLen` be used?  How do we decide on `ts_max` and `packet_max`?

j0sh:
See here for the CheckMediaLen integration https://github.com/livepeer/go-livepeer/pull/351

We hard-code an assumption of 4 seconds / 60fps segments for now, until we can configure those options during a job negotiation stage. There's a 25% allowance in goclient, so in practice it'll cut off after 5 seconds or 300 packets (a 'packet' doesn't necessarily correspond to an entire frame, but close enough for our purposes; otherwise we'd have to decode).

ericxtang:
Sounds good.  🚢

PR #72, opened 2018-04-27
Small Cleanup

ericxtang:
Additional control to start RTMP or HTTP (for example, no reason for bootnode to run RTMP server).

### DIFF
--- a/segmenter/video_segmenter.go
+++ b/segmenter/video_segmenter.go
@@ -15,7 +15,6 @@ import (
 
 	"github.com/ericxtang/m3u8"
 	"github.com/golang/glog"
-	"github.com/livepeer/go-livepeer/common"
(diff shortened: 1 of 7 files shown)


j0sh:
Looks good! 🚢 

Tests are breaking but I think that's unrelated -- maybe to the install script change from last night making the ffmpeg/ffprobe executables unavailable to tests. Will check it out later.

segmenter/video_segmenter.go
```diff
@@ -15,6 +15,7 @@ import (
 
 	"github.com/ericxtang/m3u8"
 	"github.com/golang/glog"
+	"github.com/livepeer/go-livepeer/common"
```
j0sh: Do we want to make goclient a circular dependency?

PR #74, opened 2018-05-01
Use a LPMS config struct, add RTMP/HTTP binding IPs.

j0sh:
The defaults listen to all interfaces, so we need to pass in localhost explicitly from goclient.


ericxtang:
Seems like circle needs updated grep?  https://circleci.com/gh/livepeer/lpms/163

PR #90, opened 2018-09-18
Retry Segmenter in case of disconnect.

RanjeetKaur17:
1. Added retry for RTMPToHLS
2. Added retry for PollSegment


core/lpms.go
```diff
@@ -131,6 +133,19 @@ func (l *LPMS) HandleHLSPlay(
 	l.vidPlayer.HandleHLSPlay(getMasterPlaylist, getMediaPlaylist, getSegment)
 }
 
+func (l *LPMS) RTMPToHLS(s *segmenter.FFMpegVideoSegmenter, ctx context.Context, cleanup bool) error{
+	var err error
+	for i:=0; i < RETRY_COUNT; i++ {
+		err = s.RTMPToHLS(ctx, true)
```
j0sh: presumably `cleanup` was meant here, rather than `true` ?

core/lpms.go
```diff
@@ -131,6 +133,19 @@ func (l *LPMS) HandleHLSPlay(
...
+		if err == nil {
+			break
+		} else if i < RETRY_COUNT {
+			glog.Errorf("Error Invoking Segmenter: %v, Retrying", err)
```
j0sh: We normally use a [backoff module](https://github.com/cenkalti/backoff) but maybe here it's OK to put a small sleep before retrying, say 500ms.

core/lpms.go
```diff
@@ -131,6 +133,19 @@ func (l *LPMS) HandleHLSPlay(
 	l.vidPlayer.HandleHLSPlay(getMasterPlaylist, getMediaPlaylist, getSegment)
 }
 
+func (l *LPMS) RTMPToHLS(s *segmenter.FFMpegVideoSegmenter, ctx context.Context, cleanup bool) error{
+	var err error
+	for i:=0; i < RETRY_COUNT; i++ {
```
j0sh: Ideally we'd continue until the RTMP stream is terminates but this isn't so easily done since the `RtmpVideoStream` interface doesn't expose the information we need. Fixing that is probably outside the scope of this bounty, so this is probably OK for now.

segmenter/video_segmenter_test.go
```diff
@@ -529,3 +529,25 @@ func TestMissingKeyframe(t *testing.T) {
 		return
 	}
 }
+
+func TestRetryRTMPToHLS(t *testing.T) {
```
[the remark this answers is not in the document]
RanjeetKaur17: Please ignore this test case. I have added a proper test case in lpms_test.go, please review that.

core/lpms.go
```diff
@@ -159,10 +174,16 @@ func (l *LPMS) SegmentRTMPToHLS(ctx context.Context, rs stream.RTMPVideoStream,
...
+
+				for i:=0; i < RETRY_COUNT; i++ {
+					seg, err = s.PollSegment(segCtx)
+					if err == nil {
+						break
+					} else if i < RETRY_COUNT {
+						glog.Errorf("Error polling Segment: %v, Retrying", err)
+					}
 				}
+
```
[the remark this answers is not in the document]
RanjeetKaur17: Added Handling

core/lpms.go
RanjeetKaur17: Added 500ms Sleep

core/lpms.go
```diff
@@ -159,10 +174,16 @@ func (l *LPMS) SegmentRTMPToHLS(ctx context.Context, rs stream.RTMPVideoStream,
 					glog.Errorf("HLS Stream is nil")
 					return segmenter.ErrSegmenter
 				}
-				seg, err = s.PollSegment(segCtx)
-				if err != nil {
-					return err
+
+				for i:=0; i < RETRY_COUNT; i++ {
+					seg, err = s.PollSegment(segCtx)
```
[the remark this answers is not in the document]
RanjeetKaur17: Breaking for "context.Canceled" and "context.DeadlineExceeded". Please advise if we should break on any other type of error.

core/lpms.go
```diff
@@ -131,6 +135,20 @@ func (l *LPMS) HandleHLSPlay(
 	l.vidPlayer.HandleHLSPlay(getMasterPlaylist, getMediaPlaylist, getSegment)
 }
 
+func (l *LPMS) RTMPToHLS(s *segmenter.FFMpegVideoSegmenter, ctx context.Context, cleanup bool) error{
```
j0sh: Since this doesn't use any instance members and is internal so far, maybe we can change the function to be static and/or even private. Then to make testing easier, pass in the `segmenter.VideoSegmenter` interface, eg `func rtmpToHLS(ctx context.Context, s segmenter.VideoSegmenter, cleanup bool)`

See the note down below about testing changes.

core/lpms_test.go
```diff
@@ -0,0 +1,109 @@
...
+}
+func (s *TestStream) WriteHLSPlaylistToStream(pl m3u8.MediaPlaylist) error                { return nil }
+func (s *TestStream) WriteHLSSegmentToStream(seg stream.HLSSegment) error                 { return nil }
+func (s *TestStream) ReadHLSFromStream(ctx context.Context, buffer stream.HLSMuxer) error { return nil }
+func (s *TestStream) ReadHLSSegment() (stream.HLSSegment, error)                          { return stream.HLSSegment{}, nil }
+func (s *TestStream) Width() int                                                          { return 0 }
+func (s *TestStream) Height() int                                                         { return 0 }
+func (s *TestStream) Close()															  {}
+
+func TestRetryRTMPToHLS(t *testing.T) {
```
j0sh: Thanks for the work on testing! We can probably simplify this substantially by stubbing out the segmenter. That way, we can avoid instantiating so much networking and file IO.

This also allows us to ensure that the retries are actually being called, by counting the invocations to the stubbed `segmenter.RTMPToHLS`.  Note that the `segmenter.VideoSegmenter` interface (currently empty) will need the RTMPToHLS function added to it.

Does that make sense?

core/lpms.go
```diff
@@ -159,10 +177,23 @@ func (l *LPMS) SegmentRTMPToHLS(ctx context.Context, rs stream.RTMPVideoStream,
 					glog.Errorf("HLS Stream is nil")
 					return segmenter.ErrSegmenter
 				}
-				seg, err = s.PollSegment(segCtx)
+
+				for i:=0; i < RetryCount; i++ {
+					seg, err = s.PollSegment(segCtx)
+					if err == nil {
+						break
+					} else if err == context.Canceled || err == context.DeadlineExceeded {
```
j0sh: this could probably be combined with `err == nil || ...`

core/lpms.go
RanjeetKaur17: Done

core/lpms_test.go
RanjeetKaur17: Do u mean creating a test case for rtmpToHLS static method, with a TestVideoSegmenter. Fail it first two times and verify that it is being attempted thrice?

I have made changes as per that, but I am not sure if this is what you meant. Please confirm.

core/lpms.go
RanjeetKaur17: Done

core/lpms_test.go
```diff
@@ -0,0 +1,35 @@
...
+	return nil
+}
+
+func (t *TestVideoSegmenter) GetCount() int {
+	return t.count
+}
+
+func TestRetryRTMPToHLS(t *testing.T) {
+	var testVideoSegmenter = &TestVideoSegmenter{}
+	ctx, _ := context.WithTimeout(context.Background(), time.Second*5)
```
j0sh: Could probably reset `SegmenterRetryWait` here to 1ms (or even 0?) to make the test run faster

RanjeetKaur17: Done

core/lpms_test.go
```diff
@@ -0,0 +1,35 @@
...
+	rtmpToHLS(testVideoSegmenter, ctx, true)
+	count := testVideoSegmenter.GetCount()
+	if count != 3 {
```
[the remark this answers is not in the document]
RanjeetKaur17: Miss on my part. Changed now.

core/lpms_test.go
```diff
@@ -0,0 +1,35 @@
...
+	"time"
+)
+
+type TestVideoSegmenter struct {
+	count int
+}
+
+func (t *TestVideoSegmenter) RTMPToHLS(ctx context.Context, cleanup bool) error {
+	t.count++
+	if t.count < 3 {
```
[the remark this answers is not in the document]
RanjeetKaur17: Miss on my part. Changed now.

j0sh:
Squashed this PR locally, ran through `go fmt` and a few small changes and pushed to master. Thanks for doing this!

PR #94, opened 2019-02-12
ffmpeg: Drop out-of-order packets

j0sh:
This occurs with ManyCam for Android when the device is not able to
    keep up with processing the video. Rather than deal with broken
    timestamps, we just drop the packet. https://github.com/livepeer/lpms/commit/bb48081d74f63013fad0d6279c49b59fe8f13147

ffmpeg: Remove the length checker. https://github.com/livepeer/lpms/commit/af5e976e7a9b3baad7765856b2870705779b8850


j0sh:
Fixed up some comments in the test cases and merged

PR #97, opened 2019-02-25
Close connection if stream id was not created

darkdarkdragon:
Right now, if `makeStreamID` returns empty string, connection just left hanging, so any any client (FFmpeg, OBS) just hangs trying to send data and not getting any response.

### DIFF
--- a/vidlistener/listener.go
+++ b/vidlistener/listener.go
@@ -37,6 +37,7 @@ func (self *VidListener) HandleRTMPPublish(
 
 			strmID := makeStreamID(conn.URL)
 			if strmID == "" {
+				conn.Close()
 				return
 			}
 			s := stream.NewBasicRTMPVideoStream(strmID)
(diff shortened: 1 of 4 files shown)


vidlistener/listener.go
```diff
@@ -37,6 +37,8 @@ func (self *VidListener) HandleRTMPPublish(
 
 			strmID := makeStreamID(conn.URL)
 			if strmID == "" {
+				glog.Errorf("Error RTMP makeStreamID handler - empty stream ID")
```
j0sh: The error should be logged by the `makeStreamID` caller so I don't think we need this here

darkdarkdragon: Removed

darkdarkdragon:
Added test

j0sh:
LGTM. Squashed a fixup commit locally and rebased onto master.

PR #110, opened 2019-03-11
ffmpeg: Remove one-thread-per-segment restriction.

j0sh:
This change allows ffmpeg to use as many threads as it can for
the encoding process. The renditions for each frame are still
processed sequentially, but each rendition can now use multiple
threads.

This does make it more difficult to characterize the overall behavior
or capacity of a node, but it is necessary to go fast for well-tuned
nodes.

Fixes https://github.com/livepeer/go-livepeer/issues/788

### DIFF
--- a/ffmpeg/lpms_ffmpeg.c
+++ b/ffmpeg/lpms_ffmpeg.c
@@ -203,7 +203,6 @@ static int open_output(struct output_ctx *octx, struct input_ctx *ictx)
     if (octx->fps.den) vc->framerate = av_buffersink_get_frame_rate(octx->vf.sink_ctx);
     if (octx->fps.den) vc->time_base = av_inv_q(octx->fps);
     if (octx->bitrate) vc->rc_min_rate = vc->rc_max_rate = vc->rc_buffer_size = octx->bitrate;
-    vc->thread_count = 1;
     vc->pix_fmt = av_buffersink_get_format(octx->vf.sink_ctx); // XXX select based on encoder + input support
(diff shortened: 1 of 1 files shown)


iameli:
I'ma build this against the demo UX and see what pops out!

iameli:
Definitely a huge improvement. http://media-staging.livepeer.org.s3-website.us-east-2.amazonaws.com/demo

It's interesting to watch the CPU usage on these machines now

iameli:
Definitely an enormouse improvement. http://media-staging.livepeer.org.s3-website.us-east-2.amazonaws.com/demo

It's interesting to watch the CPU usage on these machines now — processing is much more distributed but it still kinda looks like there's a single-core bottleneck at some point. Curious to know what that might be... but yeah, this looks great.

j0sh:
There's a few things going on that would introduce a sequential dependency:

* Each rendition is processed one-by-one. So with 1080, 720, 480, we'd be encoding 1080, then 720, then 480. However, from the [benchmarks](https://github.com/livepeer/go-livepeer/issues/788), this may not necessarily lead to faster encodes overall.

* Segments are very short (2 seconds), and we're setting up / tearing down everything for each segment, so we don't have much of an opportunity to amortize that overhead. We could try to keep contexts open for as long as the session is alive on the orchestrator, but that's a significant refactoring over the current flow.

At some point I'll benchmark this against GPU as well; it will also be interesting to see how this works with GPU, especially the setup / teardown overhead.

PR #114, opened 2019-05-01
Preliminary nvidia support.

j0sh:
This patch set introduces preliminary support for processing on nvidia GPUs. 

* Decode
* Encode
* Rescale
* Selectable GPUs

Various combinations are supported:

* Software-only processing
* Software decode, GPU rescale, GPU encode
* GPU decode, GPU rescale, software encode
* GPU-only processing

As much processing is kept on the GPU device as possible. Eg, if rescaling is required while only one of encode / decode is run on the GPU, then the rescaling will be done on the GPU. Only one GPU is used at a time for a given transcode.

A new API is introduced to support GPU : `Transcode2` . The old `Transcode` API has been internally retrofitted to use the new API with software-only processing, but to take advantage of GPUs, the new API will need to be used. (https://github.com/livepeer/lpms/commit/19dd8cb49ca136520596beb0cd9157dc6cfdc28c). We can remove the old `Transcode` API after the new `Transcode2` has been fully integrated into the goclient.

Sample program : https://github.com/livepeer/lpms/commit/890aab5bc160122ae625a22c47adec255ec48ca4

There are some test cases for GPU processing, however they require:
* Proper hardware
* Run with the `-tag=nvidia` flag
* Commit: https://github.com/livepeer/lpms/commit/36aa9dfe2a3dadd8a6ba190f6d79dacb28120c65

See the README for further details.

### DIFF
--- a/ffmpeg/nvidia_test.go
+++ b/ffmpeg/nvidia_test.go
@@ -1,0 +1,294 @@
+// +build nvidia
+
+package ffmpeg
+
+import (
+	"fmt"
+	"os"
+	"testing"
+)
+
+func TestNvidia_Transcoding(t *testing.T) {
+	// Various Nvidia GPU tests for encoding + decoding
+	// XXX what is missing is a way to verify these are *actually* running on GPU!
+
+	run, dir := setupTest(t)
+	defer os.RemoveAll(dir)
+
+	cmd := `
+    set -eux
+    cd "$0"
+
+    # set up initial input; truncate test.ts file
+    ffmpeg -loglevel warning -i "$1"/../transcoder/test.ts -c:a copy -c:v copy -t 1 test.ts
+  `
+	run(cmd)
+
+	var err error
+	fname := dir + "/test.ts"
+	oname := dir + "/out.ts"
+	prof := P240p30fps16x9
+
+	// hw dec, sw enc
+	err = Transcode2(&TranscodeOptionsIn{
+		Fname: fname,
+		Accel: Nvidia,
+	}, []TranscodeOptions{
+		TranscodeOptions{
+			Oname:   oname,
+			Profile: prof,
+			Accel:   Software,
+		},
+	})
+	if err != nil {
+		t.Error(err)
+	}
+
+	// sw dec, hw enc
+	err = Transcode2(&TranscodeOptionsIn{
+		Fname: fname,
+		Accel: Software,
+	}, []TranscodeOptions{
+		TranscodeOptions{
+			Oname:   oname,
+			Profile: prof,
+			Accel:   Nvidia,
+		},
+	})
+	if err != nil {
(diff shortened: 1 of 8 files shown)


j0sh:
Made a small change to add `+1` to extra_hw_frames based on feedback from @lu-zero

https://github.com/livepeer/lpms/pull/114/files#diff-5e694afe073853397a89ba4797aad0d6R364

j0sh:
Updated with latest fixes and generally fixed up with lots more tests.

ffmpeg/nvidia_test.go
```diff
@@ -0,0 +1,398 @@
...
+
+import (
+	"fmt"
+	"os"
+	"testing"
+)
+
+func TestNvidia_Transcoding(t *testing.T) {
+	// Various Nvidia GPU tests for encoding + decoding
+	// XXX what is missing is a way to verify these are *actually* running on GPU!
```
darkdarkdragon: If it's encoding in software, then it output always will be bit-by-bit exact? And with hardware it will be different?
So probably we can save hash of output made by software codec and check if it is different?

j0sh: > If it's encoding in software, then it output always will be bit-by-bit exact?

I'm not sure if we can depend on software encoding being bitexact anymore since https://github.com/livepeer/lpms/commit/a12c7a342702837229ede660e1500c1912420b4d

> So probably we can save hash of output made by software codec and check if it is different?

It'd tell us that "this is not encoded with x264" but does not actually ensure it was run on the GPU (or the 'correct' GPU as we select specific GPUs in some other tests).

I'm not sure how practical / feasible it is to check this via unit tests, but would love some other ideas. Parsing `nvidia-smi` output (super racy)?

darkdarkdragon: >It'd tell us that "this is not encoded with x264" but does not actually ensure it was run on the GPU (or the 'correct' GPU as we select specific GPUs in some other tests).

We're explicitly asking for `h264_nvenc`. If it isn't available will `avlib` return error? Then we should just believe `avlib` that it used GPU indeed.
I don't see any (reliable) way to ensure that GPU (and which one) was used.

darkdarkdragon:
LGTM

ffmpeg/nvidia_test.go
j0sh: > If it isn't available will avlib return error? 

Yes, it should.

> Then we should just believe avlib that it used GPU indeed.

The idea is to catch the possibility where a programming error (in the future) causes the GPU to *not* be used when it should have. (And vice versa.) That'd be a regression, and it's valuable to have tests catch that.

> I don't see any (reliable) way to ensure that GPU (and which one) was used.

Agreed, it's tricky. 🤔

PR #122, opened 2019-05-22
WIP: Configurable muxer options

j0sh:
Works nicely as-is, but not of much use until either of the following is true: 

* Additional container / format support becomes a product requirement (eg, fMP4 or DASH)
* Stream copy is fully implemented within the transcoder, which allows us to subsume the standalone segmenter with a combination of mux options + streamcopy. We need stream copy to copy audio anyway.

**Requires the GPU branch**

### DIFF
--- a/ffmpeg/ffmpeg.go
+++ b/ffmpeg/ffmpeg.go
@@ -39,6 +39,8 @@ type TranscodeOptions struct {
 	Profile VideoProfile
 	Accel   Acceleration
 	Device  string
+	Muxer   string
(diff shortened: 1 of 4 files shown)


j0sh:
One area where specifying a muxer is sort of useful right now : for benchmarking, `Muxer: "null", Oname: "-"` avoids disk IO, eg in https://gist.github.com/j0sh/640cadfe39b759c38f953baca11e78b0

j0sh:
Superseded by https://github.com/livepeer/lpms/pull/147

PR #123, opened 2019-05-22
vidlistener: Support opaque user data.

j0sh:
This allows us to pass around per-stream data from MakeStreamID
without storing it outside LPMS.

In support of https://github.com/livepeer/go-livepeer/issues/793

### DIFF
--- a/stream/basic_rtmp_videostream.go
+++ b/stream/basic_rtmp_videostream.go
@@ -13,7 +13,7 @@ import (
 )
 
 type BasicRTMPVideoStream struct {
-	streamID     string
+	appData      AppData // opaque app-supplied data
 	ch           chan *av.Packet
 	listeners    map[string]av.MuxCloser
 	listnersLock *sync.Mutex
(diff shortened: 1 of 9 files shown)


stream/interface.go
```diff
@@ -7,7 +7,12 @@ import (
 	"github.com/nareix/joy4/av"
 )
 
+type AppData interface {
+	StreamID() string
```
angyangie: Do we implement SteamID() anywhere?

j0sh: Yes, it's part of the interface supplied by callers of the API

goclient:

https://github.com/livepeer/go-livepeer/blob/05a5bae7161810ff3c9a9292e18d574080f36f98/server/mediaserver.go#L62-L70

testing: 

https://github.com/livepeer/lpms/blob/b845e8c33f11f4f355e8dbd5564bc0e8ce851022/stream/basic_rtmp_videostream_test.go#L55-L59

j0sh: Using `StreamID()` as part of the interface here does two things here:

* Shims into the old `GetStreamID` call . I'm not sure if we can really remove this entirely yet, some other places in LPMS code use a notion of "stream ID" but the stream interfaces in general are probably due for closer review and cleaning.
 
* Allows us to avoid defining an empty interface which is a bit of a golang smell

angyangie: Interesting! That makes sense. So we're defining the interface in lpms but only implementing it in go-livepeer? Is that problematic at all?

j0sh: > Is that problematic at all?

Not at all; it's exactly how interfaces are meant to be used: to separate definition and implementation.

Go's flavor of subtyping also allows you to twist interfaces on their head a bit by allowing you to define an interface *where you need one* on top of an existing implementation. So, rather than write an implementation following an interface, you write the interface following the implementation. A bit duck typing-ish in that way, but that's not really relevant here.

j0sh:
Fixed tests

stream/interface.go
angyangie: Interesting on writing the interface following the implementation! This makes sense as a core feature of go

stream/basic_rtmp_videostream.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 type BasicRTMPVideoStream struct {
-	streamID     string
+	opaque       AppData // opaque app-supplied data
```
yondonfu: Consider naming this `appData`? IMO it isn't that clear what `opaque` is when reading the rest of the code

j0sh: Fixed

Issue #127, opened 2019-06-20
m3u8 input broken


j0sh:
Currently, the transcoder does not behave well with m3u8 input files. This can be verified by passing a m3u8 in as input to the sample program. https://github.com/livepeer/lpms/blob/master/cmd/transcoding/transcoding.go

j0sh:
Fixed by https://github.com/livepeer/lpms/pull/131

PR #131, opened 2019-07-02
Drain queued frames from the filter and encoder.

j0sh:
https://github.com/livepeer/lpms/commit/4e8c118536ac7cd69de174dfa3f653e5ef4974f6 : ffmpeg: Drain queued frames from the filter and encoder.  …
This allows for better resource management.

Multiple frames may be output for a given filter call, such as
upsampling the frame rate. This led to running out of
surfaces in GPU-encoding mode, and steadily increasing memory usage
in software-encoding mode.

Likewise, GOP structure may result in the encoder having several frames ready
to be returned after being fed earlier frames with no immediate result.

The transcode loop is reworked to:

0. Decode a frame
1. Feed the filter graph with the decoded frame
1a. Feed and drain the encoder if no filter graph is configured
2. Drain the filter graph as much as possible
3. Feed and drain the encoder for each frame received from the filter
4. Mux and output the resulting packets.

https://github.com/livepeer/lpms/commit/839b76208dc016f9095cad5e8c07168333eb3d44 : ffmpeg: Disable hardware-specific tests.  …
Seems to have worked on a P100 but breaks with a K80.

Disable for now until a more robust set of tests is in place.

Fixes https://github.com/livepeer/lpms/issues/127

### DIFF
--- a/ffmpeg/lpms_ffmpeg.c
+++ b/ffmpeg/lpms_ffmpeg.c
@@ -366,7 +366,12 @@ static int open_input(input_params *params, struct input_ctx *ctx)
       frames->sw_format = vc->pix_fmt;
       frames->width = vc->width;
       frames->height = vc->height;
-      vc->extra_hw_frames = 16 + 1; // H.264 max refs but increases mem usage
+
+      // May want to allocate extra HW frames if we encounter samples where
+      // the defaults are insufficient. Raising this increases GPU memory usage
+      // For now, the defaults seems OK.
+      //vc->extra_hw_frames = 16 + 1; // H.264 max refs
+
       ret = av_hwframe_ctx_init(vc->hw_frames_ctx);
       if (AVERROR(ENOSYS) == ret) ret = lpms_ERR_INPUT_PIXFMT; // most likely
       if (ret < 0) dd_err("Unable to initialize a hardware frame pool\n")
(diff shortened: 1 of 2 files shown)


ffmpeg/nvidia_test.go
```diff
@@ -103,6 +103,8 @@ func TestNvidia_Transcoding(t *testing.T) {
 
 func TestNvidia_Pixfmts(t *testing.T) {
 
+	return // This test only seems to work with P100
```
mk-livepeer: is it possible to test for a P100 instead of aborting the whole test?

ffmpeg/lpms_ffmpeg.c
```diff
@@ -650,68 +705,46 @@ int process_out(struct input_ctx *ictx, struct output_ctx *octx, AVCodecContext
...
-  if (AVMEDIA_TYPE_AUDIO == ost->codecpar->codec_type) {
-      if (octx->drop_ts == AV_NOPTS_VALUE) octx->drop_ts = pkt.pts;
-      if (pkt.pts && pkt.pts == octx->drop_ts) goto proc_cleanup;
+  // Sometimes we have to reset the filter if the HW context is updated
+  // because we initially set the filter before the decoder is fully ready
+  // and the decoder may change HW params
+  if (AVMEDIA_TYPE_VIDEO == ost->codecpar->codec_type &&
+      inf && inf->hw_frames_ctx && filter->hwframes &&
+      inf->hw_frames_ctx->data != filter->hwframes) {
+    free_filter(&octx->vf); // XXX really should flush filter first
```
mk-livepeer: is this some tech debt that we should write a ticket for and take care of later?   or might it be something small that can be added to this PR?

j0sh: If this is referring to the "really should flush filter first" comment -- Not entirely sure TBH. The reason I'm hesitant to fix this [1] or make a issue is mostly because I haven't yet looked at how to construct a test case to trigger breakage and verify the fix. Without a breaking case, it may not even be a legitimate concern. (I just spent a few minutes trying some things without luck.)

[1] Which should be simpler to fix now with this changeset. Most everything between the `while(1)` could be moved into a standalone function and invoked just prior to freeing the filters. But again, I don't want to blindly add code without demonstrating actual breakage.

ffmpeg/nvidia_test.go
j0sh: Probably via parsing the output of `nvidia-smi` . I'd have to reconfigure this instance with a P100 to confirm the string to use.

Ideally the test would be made more robust so we wouldn't be gating on specific cards but need to look into our options a little more closely.  Created https://github.com/livepeer/lpms/issues/134 so it isn't lost in the shuffle.

PR #140, opened 2019-07-27
[RFC] return video pixel and frame counts after transcode

### LINKED ISSUE #124: Return pixels processed in FFMpegSegmentTranscoder.Transcode()
At the moment, transcoding of a segment is kicked off by calling:

```
func (t *FFMpegSegmentTranscoder) Transcode(fname string) ([][]byte, error)
```

In order to implement livepeer/go-livepeer#896, an orchestrator needs to know the # of pixels processed during transcoding. In the short term, we just need to know the # of output pixels (we can leave input pixel accounting for the future).

We can update the function to:

```
func (t *FFMpegSegmentTranscoder) Transcode(fname string) (res [][]byte, receipts []Receipt, err error)
```

where `receipts[i]` corresponds to `res[i]`. The function will return a receipt for each transcoded result that contains the # of pixels processed for the transcoded result (perhaps other data could be included in this receipt in the future as well?).

This will require a change to `ffmpeg.Transcode` as well.

mkrufky:
* ffmpeg/lpms_ffmpeg: count pixels during transcode
* ffmpeg: Transcode3 returns pixel and frame counts

closes #124

### DIFF
--- a/ffmpeg/ffmpeg.go
+++ b/ffmpeg/ffmpeg.go
@@ -3,12 +3,13 @@ package ffmpeg
 import (
 	"errors"
 	"fmt"
-	"github.com/golang/glog"
 	"path"
 	"path/filepath"
 	"strconv"
 	"strings"
 	"unsafe"
+
+	"github.com/golang/glog"
 )
 
 // #cgo pkg-config: libavformat libavfilter libavcodec libavutil libswscale gnutls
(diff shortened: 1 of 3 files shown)


mk-livepeer:
@yondonfu  has explained that as this will satisfy the requirements as-is, the code on his side is stubbed out to expect the data on a per-rendition basis.  As there is no harm is providing the data on a per-rendition basis, and it can be made readily available here, we will proceed that way.   This is ready for review in general, but will be augmented shortly to report the data on a per-rendition basis.

ffmpeg/ffmpeg.go
```diff
@@ -120,15 +121,26 @@ func accelDeviceType(accel Acceleration) (C.enum_AVHWDeviceType, error) {
 
 func Transcode2(input *TranscodeOptionsIn, ps []TranscodeOptions) error {
 
+	_, ret := Transcode3(input, ps)
+	return ret
+}
+
+type TranscodeReceipt struct {
```
j0sh: Can this be named `TranscodeResults` rather than `TranscodeReceipt` ?

ffmpeg/ffmpeg.go
```diff
@@ -120,15 +121,26 @@ func accelDeviceType(accel Acceleration) (C.enum_AVHWDeviceType, error) {
...
+	Frames int64
+	Pixels int64
+}
+
+func Transcode3(input *TranscodeOptionsIn, ps []TranscodeOptions) (res TranscodeReceipt, err error) {
```
j0sh: Should we return `res` as a pointer rather than a value here?

ffmpeg/lpms_ffmpeg.c
```diff
@@ -751,14 +751,14 @@ int process_out(struct input_ctx *ictx, struct output_ctx *octx, AVCodecContext
 
 #define MAX_OUTPUT_SIZE 10
 
-int lpms_transcode(input_params *inp, output_params *params, int nb_outputs)
+int lpms_transcode_and_count(input_params *inp, output_params *params, int nb_outputs, int64_t *framecount, int64_t *pixelcount)
```
j0sh: Can the original `lpms_transcode` API name be retained? Don't think we need to retain back-compat with this (internal) API.

Can we have a `struct transcode_result` output param, rather than separate params for each field? That way we can add new fields without also modifying the function signature (we might have to do that for https://github.com/livepeer/go-livepeer/issues/997)

ffmpeg/lpms_ffmpeg.c
```diff
@@ -797,13 +797,15 @@ int lpms_transcode(input_params *inp, output_params *params, int nb_outputs)
...
     if (ret < 0) main_err("transcoder: Unable to open output\n");
+    area += params[i].h * params[i].w;
   }
 
   av_init_packet(&ipkt);
   dframe = av_frame_alloc();
   if (!dframe) main_err("transcoder: Unable to allocate frame\n");
 
   while (1) {
+    frames++;
```
j0sh: As far as I can tell (without actually running things and checking), this only counts the number of input frames. A single input frame may produce multiple output frames. Also, there needs to be a check for audio and video; both are carried in AVFrames, but we're only tracking video pixels right now.

ffmpeg/lpms_ffmpeg.c
```diff
@@ -797,13 +797,15 @@ int lpms_transcode(input_params *inp, output_params *params, int nb_outputs)
     }
     ret = open_output(octx, &ictx);
     if (ret < 0) main_err("transcoder: Unable to open output\n");
+    area += params[i].h * params[i].w;
```
j0sh: This is probably OK for now, but could break if we ever do resolution pass-through, because the resolution could change mid-stream.

ffmpeg/ffmpeg.go
mkrufky: We can, but this is golang and we have garbage collection.   Can you expand on your thoughts here?  Why do you want to return by pointer rather than by reference?   It's bad practice to do this in golang unless you have a good reason for it.

ffmpeg/lpms_ffmpeg.c
```diff
@@ -751,14 +751,14 @@ int process_out(struct input_ctx *ictx, struct output_ctx *octx, AVCodecContext
 
 #define MAX_OUTPUT_SIZE 10
 
-int lpms_transcode(input_params *inp, output_params *params, int nb_outputs)
+int lpms_transcode_and_count(input_params *inp, output_params *params, int nb_outputs, int *framecount, int *pixelcount)
```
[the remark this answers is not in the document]
mkrufky: We cannot keep the original lpms_transcode name because this is `cgo` , which is actually the `c` language.  `C` does not support overloading function names as is supported by `c++`

So, unfortunately, the name of the function must change here.

ffmpeg/lpms_ffmpeg.c
mkrufky: You're right, but we're currently only counting video frames.  Audio can be handled in a separate pull request - this structure was created to be purposefully extensible.

ffmpeg/lpms_ffmpeg.c
mkrufky: Right.  This will be changed to report on a per-rendition basis *before* merge :-)

ffmpeg/lpms_ffmpeg.c
mkrufky: You seem to be contradicting yourself within this one comment.   This is an internal function and we do not need to have an extensible result.  I'd rather not wrap a C structure in go cgo decoration just for this purpose.

...although I see your point, I'd just rather not.   If this is really your preference, please say it a second time and I will give in ;-) .   BTW, this comment doesnt show on all of my logins.  Strange...

ffmpeg/ffmpeg.go
mkrufky: done in 23abea821757539ba27052094946ceb5d5d01f0e

ffmpeg/lpms_ffmpeg.c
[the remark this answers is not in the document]
j0sh: We aren't overloading anything, just changing the function signature. Retaining back compat for this internal API is not necessary.

ffmpeg/ffmpeg.go
j0sh: > It's bad practice to do this in golang unless you have a good reason for it.

Can you point me to guidance describing this practice? If anything, returning a pointer is leaning more on the GC.

Does the golang compiler implicitly convert `res` to a heap allocated reference? My understanding is that it doesn't, and we would be making full copies of the struct each time it's passed or returned, which seems undesirable.

Also, returning a nullable pointer seems a little more explicit to indicate nonexistent data, rather than zero values which could inadvertently be interpreted as valid data.

ffmpeg/lpms_ffmpeg.c
j0sh: >  I'd rather not wrap a C structure in cgo decoration

Not sure what you mean here - passing in a Golang data structure to C? If so, agreed that'd be crazy.

Define the result struct in C, something like: `struct transcode_result { int64 frames, pixels; };`, then convert that to a Golang `TranscodeResult` within the `Transcode3` function.

ffmpeg/lpms_ffmpeg.c
j0sh: > Audio can be handled in a separate pull request - this structure was created to be purposefully extensible.

Audio frames are still processed within this flow, so it will currently count audio frames as video. Moreover, counting input frames is incorrect, as there isn't a 1-1 correspondence with the number of input frames and number of output frames.

Unit tests are a good idea to double check the numbers that are being produced here.

PR #141, opened 2019-08-03
LPMS bottleneck fix

j0sh:
The commits in this PR are rebased and squashed from the following WIP branch: https://github.com/livepeer/lpms/compare/ja/old-bottleneck

### Background

There is a substantial overhead cost to starting new transcode sessions on a GPU. Adding more streams and GPUs [exacerbates](https://gist.github.com/j0sh/ae9e5a97e794e364a6dfe513fa2591c2) the problem [1].  Up until now, a new GPU session has been started for each segment that's received. This has led to a massive slowdown in transcoding latency, and becomes worse with each additional concurrent stream.

To mitigate this setup overhead, we make transcode sessions persistent for each stream, so segments after the first can re-use the same session.

[1] There may be a global lock somewhere within FFmpeg or the Nvidia drivers, or it may be intrinsic.

### LPMS API changes:

Transcoding becomes stateful: a transcode session is persisted on the GPU in between segments. The caller becomes responsible for terminating session. Working with the transcoder now involves allocating the transcoder, transcoding and stopping the transcoder. The API is as follows:

```golang
func NewTranscoder() *Transcoder
func (t *Transcoder) StopTranscoder()
func (t *Transcoder) Transcode(input *TranscodeOptionsIn, ps []TranscodeOptions) (*TranscodeResults, error)
```

The `Transcoder` struct is currently [opaque](https://github.com/livepeer/lpms/pull/141/files#diff-c04000b64dc45aba226819ef9f095d97R37-R41) to outside packages - nothing within that needs to be exposed at the moment.

The signature to the `Transcode` function match the existing [`Transcode3`](https://github.com/livepeer/lpms/pull/141/files#diff-c04000b64dc45aba226819ef9f095d97L157) function signature. Existing transcoding APIs are re-defined in terms of this new API and will continue to work unmodified in older programs.

Note that this doubles down on the "one segment at a time" restriction. In order to process multiple segments for a the same job in parallel, a new session needs to be created. However, this may end up being [very slow](https://gist.github.com/j0sh/ae9e5a97e794e364a6dfe513fa2591c2). Sessions should be reused wherever possible.

The older `Transcode3` API is preserved for single-shot transcoding, and is implemented as a [wrapper](https://github.com/livepeer/lpms/pull/141/files#diff-c04000b64dc45aba226819ef9f095d97R163-R167) around the three-step API.

### Code Flow and Changes

* **Muxer, Filtergraphs and Audio:** No change from what we have now. Always recreated in between sessions. Enough code may have been moved around to obscure this fact, but the there should be no net effect on the muxer, filtergraphs and audio as a result of this PR.

* **Decoding:** Reused in between sessions. The number of streams and the indices of audio / video components are also cached in between segments in order to avoid a latency-incurring probing step. If the stream ordering changes in between segments, things will likely break. (This shouldn't happen under most circumstances, but isn't unheard of, eg in case there is a reconfiguration somewhere at the source. In which case the best thing to do there would be to create a new session, rather than reuse an existing session.)

Decoder buffering is a problem: sometimes we have completed inputting all packets from a segment, but haven't received all pending frames from the decoder. In order to flush the decoder buffer, we send dummy packet with a sentinel timestamp, keep feeding decoder until we receive a decoded frame back with the sentiel timestamp. [Discard](https://github.com/livepeer/lpms/pull/141/files#diff-5e694afe073853397a89ba4797aad0d6R1182) any frames with the sentinel timestamp. The dummy packet is taken from the [first packet](https://github.com/livepeer/lpms/pull/141/files#diff-5e694afe073853397a89ba4797aad0d6R878-R881) of each segment (and [reset](https://github.com/livepeer/lpms/pull/141/files#diff-5e694afe073853397a89ba4797aad0d6R1256) for subsequent segments).

* **Encoding**: The encoder session is kept alive in between segments *if* a hardware encoder is used. If a software encoder is used, then the encoder is reset. This fact alone makes the [set-up](https://github.com/livepeer/lpms/pull/141/files#diff-5e694afe073853397a89ba4797aad0d6R1145-R1151) and [clean-up](https://github.com/livepeer/lpms/pull/141/files#diff-5e694afe073853397a89ba4797aad0d6R1252-R1259) flows [conditional](https://github.com/livepeer/lpms/pull/141/files#diff-5e694afe073853397a89ba4797aad0d6R1132-R1138) and rather confusing.

For hardware encoding, we current don't do anything to explicitly flush the encoder in between segments. This seems OK for now - we receive all frames we're supposed to, according to our [tests](https://github.com/livepeer/lpms/pull/141/files#diff-5357b1736961b1722e5c1972859ceff7R494-R550) - but this makes me nervous that certain inputs or configurations will leave us with buffered frames (eg, possibly B-frames). (There is a solution that *might* work for flushing the Nvidia encoder but it involves modifying FFmpeg in questionable ways.)

### API Usability

Some usability shortcomings of the current API that should be mentioned. (We can opt to fix this before merging, but I haven't gotten around to it. Happy to do this sooner since it'd be an API break.)

Currently, the output renditions are fixed for the duration of the transcode session, but the user can pass in whatever they want for subsequent segments [1]. We attempt to "check" this with a [rudimentary length check](https://github.com/livepeer/lpms/pull/141/files#diff-5e694afe073853397a89ba4797aad0d6R1295-R1298) of the number of outputs, but doesn't actually enforce that the configuration of those outputs match. A better API would look something like:

```
NewTranscoder(InputOptions, OutputOptions)
Transcode(InputFilenames, OutputFilenames)
```

That would mean defining new structs for `NewTranscoder` and `Transcoder` since we shouldn't re-use or modify the [old parameter sets](https://github.com/livepeer/lpms/blob/482441456844e4d3a05d316d20d4d3dfd65d61ba/ffmpeg/ffmpeg.go#L42-L57). 

However, the in/out filenames might not be the only thing we legitimately want to pass in to the transcoder on a per-segment basis, so the parameters should be a struct.

[1] Example of poor API usage:

```golang
tr := NewTranscoder()
out1 := []TranscodeOptionsOut{ ... one set of renditions ... }
tr.Transcode(in1, out1)
// Most of this is basically ignored, apart from the output filename (which *should* vary)
// Different muxer options will be respected as well, but probably shouldn't for consistency
out2 := []TranscodeOptionsOut{ ... another set of renditions ... }
tr.Transcode(in2, out2)
tr.StopTranscoder()
```

Fixes https://github.com/livepeer/go-livepeer/issues/997
Fixes https://github.com/livepeer/go-livepeer/issues/951

### DIFF
--- a/ffmpeg/ffmpeg.go
+++ b/ffmpeg/ffmpeg.go
@@ -249,7 +249,7 @@ func Transcode3(input *TranscodeOptionsIn, ps []TranscodeOptions) (*TranscodeRes
 	}
 	ret := int(C.lpms_transcode(inp, paramsPointer, resultsPointer, C.int(len(params)), decoded))
 	if 0 != ret {
-		glog.Infof("Transcoder Return : %v\n", Strerror(ret))
+		glog.Error("Transcoder Return : ", Strerror(ret))
 		return nil, ErrorMap[ret]
 	}
 	tr := make([]MediaInfo, len(ps))
--- a/ffmpeg/ffmpeg.go
+++ b/ffmpeg/ffmpeg.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
 	"unsafe"
 )
 
@@ -19,6 +20,7 @@ import "C"
 var ErrTranscoderRes = errors.New("TranscoderInvalidResolution")
 var ErrTranscoderHw = errors.New("TranscoderInvalidHardware")
 var ErrTranscoderInp = errors.New("TranscoderInvalidInput")
+var ErrTranscoderStp = errors.New("TranscoderStopped")
 
 type Acceleration int
 
@@ -31,6 +33,12 @@ const (
 type ComponentOptions struct {
 	Name string
 	Opts map[string]string
+}
+
+type Transcoder struct {
+	handle  *C.struct_transcode_thread
+	stopped bool
+	mu      *sync.Mutex
 }
 
 type TranscodeOptionsIn struct {
@@ -155,6 +163,17 @@ func Transcode2(input *TranscodeOptionsIn, ps []TranscodeOptions) error {
 }
 
 func Transcode3(input *TranscodeOptionsIn, ps []TranscodeOptions) (*TranscodeResults, error) {
+	t := NewTranscoder()
+	defer t.StopTranscoder()
+	return t.Transcode(input, ps)
+}
+
+func (t *Transcoder) Transcode(input *TranscodeOptionsIn, ps []TranscodeOptions) (*TranscodeResults, error) {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	if t.stopped || t.handle == nil {
+		return nil, ErrTranscoderStp
+	}
 	if input == nil {
 		return nil, ErrTranscoderInp
 	}
@@ -198,7 +217,7 @@ func Transcode3(input *TranscodeOptionsIn, ps []TranscodeOptions) (*TranscodeRes
 		filters += fmt.Sprintf("%s='w=if(gte(iw,ih),%d,-2):h=if(lt(iw,ih),%d,-2)'", scale_filter, w, h)
 		if input.Accel != Software && p.Accel == Software {
 			// needed for hw dec -> hw rescale -> sw enc
-			filters = filters + ":format=yuv420p,hwdownload"
+			filters = filters + ",hwdownload,format=nv12"
 		}
 		muxOpts := C.component_opts{
 			opts: newAVOpts(p.Muxer.Opts), // don't free this bc of avformat_write_header API
@@ -206,6 +225,12 @@ func Transcode3(input *TranscodeOptionsIn, ps []TranscodeOptions) (*TranscodeRes
 		if p.Muxer.Name != "" {
 			muxOpts.name = C.CString(p.Muxer.Name)
 			defer C.free(unsafe.Pointer(muxOpts.name))
+		}
+		// Set some default encoding options
+		if len(p.VideoEncoder.Name) <= 0 && len(p.VideoEncoder.Opts) <= 0 {
+			p.VideoEncoder.Opts = map[string]string{
+				"forced-idr": "1",
+			}
 		}
 		vidOpts := C.component_opts{
 			name: C.CString(encoder),
@@ -230,13 +255,28 @@ func Transcode3(input *TranscodeOptionsIn, ps []TranscodeOptions) (*TranscodeRes
 		params[i] = C.output_params{fname: oname, fps: fps,
 			w: C.int(w), h: C.int(h), bitrate: C.int(bitrate),
 			muxer: muxOpts, audio: audioOpts, video: vidOpts, vfilters: vfilt}
+		defer func(param *C.output_params) {
+			// Work around the ownership rules:
+			// ffmpeg normally takes ownership of the following AVDictionary options
+			// However, if we don't pass these opts to ffmpeg, then we need to free
+			if param.muxer.opts != nil {
+				C.av_dict_free(&param.muxer.opts)
+			}
+			if param.audio.opts != nil {
+				C.av_dict_free(&param.audio.opts)
+			}
+			if param.video.opts != nil {
+				C.av_dict_free(&param.video.opts)
+			}
+		}(&params[i])
 	}
 	var device *C.char
 	if input.Device != "" {
 		device = C.CString(input.Device)
 		defer C.free(unsafe.Pointer(device))
 	}
-	inp := &C.input_params{fname: fname, hw_type: hw_type, device: device}
+	inp := &C.input_params{fname: fname, hw_type: hw_type, device: device,
+		handle: t.handle}
 	results := make([]C.output_results, len(ps))
 	decoded := &C.output_results{}
 	var (
@@ -249,7 +289,7 @@ func Transcode3(input *TranscodeOptionsIn, ps []TranscodeOptions) (*TranscodeRes
 	}
 	ret := int(C.lpms_transcode(inp, paramsPointer, resultsPointer, C.int(len(params)), decoded))
 	if 0 != ret {
-		glog.Error("Transcoder Return : ", Strerror(ret))
+		glog.Error("Transcoder Return : ", ErrorMap[ret])
 		return nil, ErrorMap[ret]
 	}
 	tr := make([]MediaInfo, len(ps))
@@ -264,6 +304,24 @@ func Transcode3(input *TranscodeOptionsIn, ps []TranscodeOptions) (*TranscodeRes
 		Pixels: int64(decoded.pixels),
 	}
 	return &TranscodeResults{Encoded: tr, Decoded: dec}, nil
+}
+
+func NewTranscoder() *Transcoder {
+	return &Transcoder{
+		handle: C.lpms_transcode_new(),
+		mu:     &sync.Mutex{},
+	}
+}
+
+func (t *Transcoder) StopTranscoder() {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	if t.stopped {
+		return
+	}
+	C.lpms_transcode_stop(t.handle)
+	t.handle = nil // prevent accidental reuse
+	t.stopped = true
 }
 
 func InitFFmpeg() {
--- a/ffmpeg/lpms_ffmpeg.c
+++ b/ffmpeg/lpms_ffmpeg.c
@@ -1,15 +1,21 @@
 #include "lpms_ffmpeg.h"
 
+#include <libavcodec/avcodec.h>
+
 #include <libavformat/avformat.h>
 #include <libavfilter/avfilter.h>
 #include <libavfilter/buffersink.h>
 #include <libavfilter/buffersrc.h>
 #include <libavutil/opt.h>
+#include <libavutil/pixdesc.h>
+
+#include <pthread.h>
 
 // Not great to appropriate internal API like this...
 const int lpms_ERR_INPUT_PIXFMT = FFERRTAG('I','N','P','X');
 const int lpms_ERR_FILTERS = FFERRTAG('F','L','T','R');
 const int lpms_ERR_PACKET_ONLY = FFERRTAG('P','K','O','N');
+const int lpms_ERR_OUTPUTS = FFERRTAG('O','U','T','P');
 
 //
 // Internal transcoder data structures
@@ -24,8 +30,13 @@ struct input_ctx {
   // Hardware decoding support
   AVBufferRef *hw_device_ctx;
   enum AVHWDeviceType hw_type;
+  char *device;
 
   int64_t next_pts_a, next_pts_v;
+
+  // Decoder flush
+  AVPacket *first_pkt;
+  int flushed;
 };
 
 struct filter_ctx {
@@ -50,6 +61,9 @@ struct output_ctx {
   int dv, da; // flags whether to drop video or audio
   struct filter_ctx vf, af;
 
+  // Optional hardware encoding support
+  enum AVHWDeviceType hw_type;
+
   // muxer and encoder information (name + options)
   component_opts *muxer;
   component_opts *video;
@@ -58,6 +72,19 @@ struct output_ctx {
   int64_t drop_ts;     // preroll audio ts to drop
 
   output_results  *res; // data to return for this output
+
+};
+
+#define MAX_OUTPUT_SIZE 10
+
+struct transcode_thread {
+  int initialized;
+
+  struct input_ctx ictx;
+  struct output_ctx outputs[MAX_OUTPUT_SIZE];
+
+  int nb_outputs;
+
 };
 
 void lpms_init()
@@ -185,7 +212,7 @@ static void free_output(struct output_ctx *octx)
     avformat_free_context(octx->oc);
     octx->oc = NULL;
   }
-  if (octx->vc) avcodec_free_context(&octx->vc);
+  if (octx->vc && AV_HWDEVICE_TYPE_NONE == octx->hw_type) avcodec_free_context(&octx->vc);
   if (octx->ac) avcodec_free_context(&octx->ac);
   free_filter(&octx->vf);
   free_filter(&octx->af);
@@ -203,6 +230,23 @@ static int needs_decoder(char *encoder) {
   // Checks whether the given "encoder" depends on having a decoder.
   // Do this by enumerating special cases that do *not* need encoding
   return !(is_copy(encoder) || is_drop(encoder));
+}
+
+static int is_flush_frame(AVFrame *frame)
+{
+  return -1 == frame->pts;
+}
+
+static void send_first_pkt(struct input_ctx *ictx)
+{
+  if (ictx->flushed || !ictx->first_pkt) return;
+
+  int ret = avcodec_send_packet(ictx->vc, ictx->first_pkt);
+  if (ret < 0) {
+    char errstr[AV_ERROR_MAX_STRING_SIZE];
+    av_strerror(ret, errstr, sizeof errstr);
+    fprintf(stderr, "Error sending flush packet : %s\n", errstr);
+  }
 }
 
 static enum AVPixelFormat hw2pixfmt(AVCodecContext *ctx)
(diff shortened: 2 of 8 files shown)


ffmpeg/lpms_ffmpeg.c
```diff
@@ -58,6 +72,22 @@ struct output_ctx {
   int64_t drop_ts;     // preroll audio ts to drop
 
   output_results  *res; // data to return for this output
+
+  int nb_frame; // XXX temp
```
yondonfu: Why is this marked as temp?

ffmpeg/lpms_ffmpeg.c
```diff
@@ -726,20 +945,40 @@ int encode(AVCodecContext* encoder, AVFrame *frame, struct output_ctx* octx, AVS
...
 
+  int ret = 0;
   AVPacket pkt = {0};
 
   if (AVMEDIA_TYPE_VIDEO == ost->codecpar->codec_type && frame) {
     octx->res->frames++;
     octx->res->pixels += encoder->width * encoder->height;
+    if (!octx->nb_frame) {
+      //fprintf(stderr, "Forcing IFrame\n");
+      frame->pict_type = AV_PICTURE_TYPE_I;
```
yondonfu: IIUC this forces the first frame of each segment encoded to be an I-frame? Was this not the case before?

ffmpeg/lpms_ffmpeg.c
```diff
@@ -812,6 +1051,10 @@ int process_out(struct input_ctx *ictx, struct output_ctx *octx, AVCodecContext
     } else if (ret < 0) proc_err("Error consuming the filtergraph\n");
     ret = encode(encoder, frame, octx, ost);
     av_frame_unref(frame);
+    // Ror HW we keep the encoder open so will only get EAGAIN.
```
yondonfu: What does "Ror HW" mean?

ffmpeg/lpms_ffmpeg.h
```diff
@@ -36,10 +45,16 @@ typedef struct {
 typedef struct {
     int frames;
     int64_t pixels;
+
+  // Returns
+  int ret; // return code for this segment
```
yondonfu: Looks like there is some mismatch in indentation of fields within this struct

yondonfu:
In theory, if an existing handle was initialized using a certain set of output renditions, that handle could be re-used for any stream as long as the stream is using those set of output renditions right?

Updating the transcode API so that the user can only pass in parameters that can change on a per-segment basis sounds like a good idea to me.

> Fixing the output > However, the in/out filenames might not be the only thing we legitimately want to pass in to the transcoder on a per-segment basis, so the parameters should be a struct.

Do you have anything in particular in mind or would using a struct just be for future-proofing?

ffmpeg/lpms_ffmpeg.c
```diff
@@ -592,6 +450,11 @@ static int init_audio_filters(struct input_ctx *ictx, struct output_ctx *octx,
       ictx->ac->sample_rate, ictx->ac->sample_fmt, ictx->ac->channel_layout,
       ictx->ac->channels, time_base.num, time_base.den);
 
+  // TODO set sample format and rate based on encoder support,
+  //      rather than hardcoding
+  snprintf(filters_descr, sizeof filters_descr,
```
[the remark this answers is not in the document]
j0sh: Just the latter, if/when we support a more diverse set of encoders or users require the ability to customize audio transcoding parameters in the same way they can customize for video. In fact, we don't even do audio encoding in the goclient anymore, just stream copy.

ffmpeg/ffmpeg.go
```diff
@@ -33,6 +34,11 @@ type ComponentOptions struct {
 	Opts map[string]string
 }
 
+type Transcoder struct {
+	handle  *C.struct_transcode_thread
+	stopped bool
```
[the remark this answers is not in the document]
j0sh: >  we are not making Transcoder concurrency safe

The transcoder is safe to use in a multi-threaded context. It'll just block subsequent segments until it's finished with the current segment. See the main `lpms_transcode` entry point: https://github.com/livepeer/lpms/pull/141/files#diff-5e694afe073853397a89ba4797aad0d6R1274

ffmpeg/lpms_ffmpeg.c
```diff
@@ -58,6 +72,22 @@ struct output_ctx {
   int64_t drop_ts;     // preroll audio ts to drop
 
   output_results  *res; // data to return for this output
+
+  int nb_frame; // XXX temp
+};
+
+#define MAX_OUTPUT_SIZE 10
```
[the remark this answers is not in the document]
j0sh: Yes, and not a new one  - the code was just reshuffled.

ffmpeg/ffmpeg.go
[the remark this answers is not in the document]
yondonfu: Right - `lpms_transcode` should have that covered, but I think we'd want to protect `stopped` with a mutex if we were using this `Transcoder` type in a multi-threaded context.

ffmpeg/lpms_ffmpeg.c
j0sh: Ah, because it might not be needed after all. (This would also imply that the `force-idr` AVOption isn't required either.) Checking and will follow up.

ffmpeg/lpms_ffmpeg.h
j0sh: Yep, the this field was actually unused at this point. Removed.

ffmpeg/lpms_ffmpeg.c
j0sh: Typo, s/Ror/For/ . Fixed

ffmpeg/lpms_ffmpeg.c
j0sh: If the encoder is persistent across sessions, we needed some way to tell it to insert an I-frame for the beginning of a segment. But with the [flush patch](https://github.com/livepeer/lpms/pull/141/files#diff-5e694afe073853397a89ba4797aad0d6R974-R977) [1] that was just recently re-introduced, this might not be needed anymore. Need to double check that.

[1] Doh, just noticed the almost-identical code block below. Fixed locally.

ffmpeg/lpms_ffmpeg.c
```diff
@@ -687,23 +897,32 @@ int process_in(struct input_ctx *ictx, AVFrame *frame, AVPacket *pkt)
...
     pkt->stream_index = ictx->vi; // XXX ugly?
-    if (!ret) return ret;
+    if (!ret) {
+      if (is_flush_frame(frame)) ictx->flushed = 1;
+      return ret;
+    }
   }
   if (ictx->ac) {
-    avcodec_send_packet(ictx->ac, NULL);
+    avcodec_send_packet(ictx->ac, NULL); // XXX fix
```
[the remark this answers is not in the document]
j0sh: Actually the comment outdated, since we restart the audio session for each segment rather than attempt to persist it. Removed the comment.

ffmpeg/ffmpeg.go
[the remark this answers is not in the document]
j0sh: Ah I see what you mean now - good point, had missed that. Fixed.

ffmpeg/ffmpeg.go
```diff
@@ -203,6 +218,13 @@ func Transcode3(input *TranscodeOptionsIn, ps []TranscodeOptions) (*TranscodeRes
 			muxOpts.name = C.CString(p.Muxer.Name)
 			defer C.free(unsafe.Pointer(muxOpts.name))
 		}
+		// Set some default encoding options
+		if len(p.VideoEncoder.Name) <= 0 && len(p.VideoEncoder.Opts) <= 0 {
+			p.VideoEncoder.Opts = map[string]string{
```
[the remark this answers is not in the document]
j0sh: We don't set it if the user passes in their own options. Practical scenarios for open GOP and non-intra I frames include non-segmented VOD or real time streaming where you want to take advantage of a larger range of pictures for the bitrate savings.

On checking again, nvenc doesn't actually the cgop option, so it's useless for us now. Will remove. It was a leftover from when trying to kludge x264 into doing persistent transcode sessions. We aren't anymore, so this option is no longer needed.

The `forced-idr` option *might* not be needed as well...

ffmpeg/ffmpeg.go
```diff
@@ -194,7 +209,7 @@ func Transcode3(input *TranscodeOptionsIn, ps []TranscodeOptions) (*TranscodeRes
 		filters := fmt.Sprintf("fps=%d/%d,%s='w=if(gte(iw,ih),%d,-2):h=if(lt(iw,ih),%d,-2)'", param.Framerate, 1, scale_filter, w, h)
 		if input.Accel != Software && p.Accel == Software {
 			// needed for hw dec -> hw rescale -> sw enc
-			filters = filters + ":format=yuv420p,hwdownload"
+			filters = filters + ",hwdownload,format=nv12"
```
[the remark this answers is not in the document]
j0sh: Old way was actually broken - it only worked with the `npp` scaler which we phased out in favor of CUDA before the original GPU PR was even merged. See https://github.com/livepeer/go-livepeer/issues/951

This seems to work better now that we're using cuvid as a decoder, but I have to check that it actually fixes https://github.com/livepeer/go-livepeer/issues/951

ffmpeg/lpms_ffmpeg.c
```diff
@@ -670,6 +875,11 @@ int process_in(struct input_ctx *ictx, AVFrame *frame, AVPacket *pkt)
     else if (pkt->stream_index == ictx->vi || pkt->stream_index == ictx->ai) break;
     else dec_err("Could not find decoder or stream\n");
```
darkdarkdragon: Does this means that we're always doing decoding, even for stream copy?

ffmpeg/lpms_ffmpeg.c
```diff
@@ -687,23 +897,32 @@ int process_in(struct input_ctx *ictx, AVFrame *frame, AVPacket *pkt)
...
     ret = avcodec_receive_frame(ictx->vc, frame);
-    pkt->stream_index = ictx->vi; // XXX ugly?
-    if (!ret) return ret;
+    pkt->stream_index = ictx->vi;
+    if (!ret) {
+      if (is_flush_frame(frame)) ictx->flushed = 1;
+      return ret;
+    }
   }
   if (ictx->ac) {
```
darkdarkdragon: Can you describe logic here? It looks like we're getting here only if there is no video stream, or if there was error in `avcodec_receive_frame`. But why?

ffmpeg/lpms_ffmpeg.c
```diff
@@ -944,42 +1233,100 @@ int lpms_transcode(input_params *inp, output_params *params,
...
+  if (ictx->vc && AV_HWDEVICE_TYPE_NONE == ictx->hw_type) avcodec_free_context(&ictx->vc);
+  for (i = 0; i < nb_outputs; i++) free_output(&outputs[i]);
   return ret == AVERROR_EOF ? 0 : ret;
 #undef main_err
 }
+
+int lpms_transcode(input_params *inp, output_params *params,
+  output_results *results, int nb_outputs, output_results *decoded_results)
+{
+  int ret = 0;
```
darkdarkdragon: Can't this potentially overwrite value that is set inside lock down there?
(if called from another thread?)

ffmpeg/lpms_ffmpeg.c
```diff
@@ -944,42 +1233,100 @@ int lpms_transcode(input_params *inp, output_params *params,
...
+
+  if (!handle) return;
+
+  pthread_mutex_lock(&handle->mu);
+  free_input(&handle->ictx);
+  for (i = 0; i < MAX_OUTPUT_SIZE; i++) {
+    free_output(&handle->outputs[i]);
+    if (handle->outputs[i].vc) avcodec_free_context(&handle->outputs[i].vc);
+  }
+  pthread_mutex_unlock(&handle->mu);
```
darkdarkdragon: Shouldn't be `pthread_mutex_destroy` called here?

ffmpeg/lpms_ffmpeg.c
```diff
@@ -944,42 +1233,100 @@ int lpms_transcode(input_params *inp, output_params *params,
...
+
+  free(handle);
```
darkdarkdragon: If `lpms_transcode_stop` called from two threads, isn't this will lead to `free_input(&handle->ictx);` being called with pointer to uninitialized memory?

j0sh:
> So we will skip B-frames in case of stream copy?

@darkdarkdragon Damn! Have to test for this, but you're probably right. Good catch!

ffmpeg/lpms_ffmpeg.c
j0sh: That's basically correct. Note that the "error" case also indicates normal cases where there's no video frames available from the decoder - this is also signaled by the return code, eg EAGAIN or EOF.

This is just the way we flush the decoder. If there's a nonzero response type, we know there's no video frame to return. So we continue on to the audio decoder, and attempt to return an audio frame from there.

ffmpeg/lpms_ffmpeg.c
j0sh: Stream copy does not decode if all other outputs for that stream are also copy (or drop). 

https://github.com/livepeer/lpms/blob/8d03c6b7397deaed1a2bf3cb2bafdabe0bcaa79e/ffmpeg/lpms_ffmpeg.c#L232-L236

https://github.com/livepeer/lpms/blob/8d03c6b7397deaed1a2bf3cb2bafdabe0bcaa79e/ffmpeg/lpms_ffmpeg.c#L1279-L1283

(Note there's a little trick here used to force decoding if there's no outputs - we depend on that when checking pixel counts!)

If all the outputs of a given stream are copy (or drop), then we shouldn't decode - the flow will hit line 875 (right above this line) to check the stream index, and exits early from the decode loop.

https://github.com/livepeer/lpms/blob/8d03c6b7397deaed1a2bf3cb2bafdabe0bcaa79e/ffmpeg/lpms_ffmpeg.c#L875

ffmpeg/lpms_ffmpeg.c
j0sh: Hmm, not sure I understand?

`ret` is function-scoped so another thread calling the same function would initialize a new copy of `ret`.

ffmpeg/lpms_ffmpeg.c
j0sh: Indeed. Combined with https://github.com/livepeer/lpms/pull/141#discussion_r348263123 and https://github.com/livepeer/lpms/pull/141#discussion_r348028335 , it seems much better to just handle the locking within golang itself. Having a GC helps. Fixed in https://github.com/livepeer/lpms/pull/141/commits/39fe9c5cac114fa0f3ddacdc5c22270f40b65db8

ffmpeg/lpms_ffmpeg.c
j0sh: Yeah - `pthread_mutex_init` doesn't actually allocate so this never showed up as a problem during testing - but you're right, it should have been called. Fixed by replacing C mutexes with golang. Also see  https://github.com/livepeer/lpms/pull/141#discussion_r349382363

 https://github.com/livepeer/lpms/pull/141/commits/39fe9c5cac114fa0f3ddacdc5c22270f40b65db8

ffmpeg/lpms_ffmpeg.c
darkdarkdragon: My mistake, haven't touched threading in C for a while...

j0sh:
> Updating the transcode API so that the user can only pass in parameters that can change on a per-segment basis sounds like a good idea to me.

Opted not to do this for now - both because the actual API itself requires some more thought and discussion that I'd like to have some more time for, and because any such API would likely have knock-on effects on the go-livepeer integration that would require even more work there.

Since there isn't really an appetite for prolonging this PR even further, we can keep the API as-is for now.

cc @yondonfu

ffmpeg/lpms_ffmpeg.c
j0sh: This turns out to have been necessary even with the flush patch, so we keep it in, but in slightly modified form using `octx->res->frames` which already tracks the output frame count for the segment.

ffmpeg/lpms_ffmpeg.c
j0sh: Keeping track of the frame count is still necessary, but we already have that information for the other pixel counts, so we rely on `res.frames` rather than using this separate counter. Removed `nb_frame` from the struct.

darkdarkdragon:
@j0sh what about my [comment](https://github.com/livepeer/lpms/pull/141#pullrequestreview-319469634) about skipping B-frames?

ffmpeg/lpms_ffmpeg.c
darkdarkdragon: Thanks! Maybe add comment explaining this to the code?

ffmpeg/lpms_ffmpeg.c
darkdarkdragon: Thanks!

darkdarkdragon:
@j0sh 
I'm missing something, or this is 
https://github.com/livepeer/lpms/pull/141/files#diff-5e694afe073853397a89ba4797aad0d6L662-L664
```c
  av_init_packet(pkt);
  // loop until a new frame has been decoded, or EAGAIN
  while (1) {
```
not a loop anymore? It has unconditional break at the end?

j0sh:
> @j0sh what about my comment about skipping B-frames?

Not in this PR, since that's an unrelated issue.

> this is not a loop anymore? It has unconditional break at the end?

Ah yeah - this PR didn't change anything there, but stream copy changed the behavior so the comment has been outdated. Updated the comment, but avoided code changes in https://github.com/livepeer/lpms/pull/141/commits/f1f732d64a5c82fd2bdb67ae5d734b77d1d49b7e

ffmpeg/lpms_ffmpeg.c
j0sh: Comment added in https://github.com/livepeer/lpms/pull/141/commits/f7fb9b6ea5cbcd08e760a40c0e03470f4376ebd6

darkdarkdragon:
@j0sh 
> Not in this PR, since that's an unrelated issue.

do we have issue created in Github for this?

otherwise, LGTM, 🚢 !

ffmpeg/ffmpeg.go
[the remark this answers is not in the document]
yondonfu: So `forced-idr` is in fact needed here?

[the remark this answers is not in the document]
j0sh: Yes the `force-idr` option is required - without it, we get decoder errors at the beginning of playback because SPS/PPS extradata isn't always is emitted for the first frame of new segments, which we signal via an explicit keyframe. This option forces that.

Ideally this option wouldn't be needed at all; ffmpeg itself *should* but smart enough to emit extradata for the first frame after a flush; I may add that in.

Also just checked and it seems that disabling this option *doesn't* lead to any unit test failures - we really should capture this behavior in tests. Will make a note to correct that.

j0sh:
> do we have issue created in Github for this?

Tracking here: https://github.com/livepeer/lpms/issues/164

> we really should capture this behavior in tests. Will make a note to correct that.

Tracking here: https://github.com/livepeer/lpms/issues/165

Issue #142, opened 2019-08-08
stream/basic_hls_videostream.go:90:11: s.plCache.AppendSegment undefined


mkrufky:
This is currently a blocker on getting `lpms` to play nicely with `go mod`
```
stream/basic_hls_videostream.go:90:11: s.plCache.AppendSegment undefined (type *m3u8.MediaPlaylist has no field or method AppendSegment)
```

PR #143, opened 2019-08-08
move to github.com/livepeer/m3u8 v0.11.0

mk-livepeer:
update github.com/livepeer/m3u8 dependency to v0.11.0


go.sum
```diff
@@ -1,6 +1,12 @@
+github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
```
j0sh: Curious, where does this come from? Is this file the transitive closure of dependencies for LPMS -> m3u8 -> testify -> (others) ?

j0sh:
This PR fixes https://github.com/livepeer/lpms/issues/142

go.sum
mkrufky: I believe so

Issue #137, opened 2019-07-10
go tests fail: `TestNoRTMPListener` and `TestServerDisconnectMidStream`


mk-livepeer:
blocks #136
```$ go test ./...
?   	github.com/livepeer/lpms/cmd/example	[no test files]
?   	github.com/livepeer/lpms/cmd/example/simple	[no test files]
?   	github.com/livepeer/lpms/cmd/transcoding	[no test files]
ok  	github.com/livepeer/lpms/core	(cached)
ok  	github.com/livepeer/lpms/ffmpeg	(cached)
HandShake: client signature does not match!
[hls @ 0x901ee00] Cannot use rename on non file protocol, this may lead to races and temporary partial files
WriteN, RTMP send error 32 (5 bytes)
WriteN, RTMP send error 32 (42 bytes)
RTMP_ReadPacket, failed to read RTMP packet body. len: 5604
HandShake: client signature does not match!
WriteN, RTMP send error 32 (5 bytes)
WriteN, RTMP send error 32 (42 bytes)
RTMP_ReadPacket, failed to read RTMP packet body. len: 5604
No application or playpath in URL!
RTMP_Connect0, failed to connect socket. 61 (Connection refused)
segmenter: Unable to open input
--- FAIL: TestNoRTMPListener (0.00s)
    video_segmenter_test.go:419: error was not nil; got  Unknown error occurred
HandShake: client signature does not match!
RTMP_ReadPacket, failed to read RTMP packet header
[flv @ 0xa002400] Could not find codec parameters for stream 0 (Video: h264, none): unspecified size
Consider increasing the value for the 'analyzeduration' and 'probesize' options
--- FAIL: TestServerDisconnectMidStream (0.11s)
    video_segmenter_test.go:464: Expected 'Input/output error' but instead got  <nil>
FAIL
FAIL	github.com/livepeer/lpms/segmenter	3.606s
ok  	github.com/livepeer/lpms/stream	(cached)
ok  	github.com/livepeer/lpms/transcoder	(cached)
ok  	github.com/livepeer/lpms/vidlistener	(cached)
ok  	github.com/livepeer/lpms/vidplayer	(cached)
```

j0sh:
[...]

Issue #136, opened 2019-07-10
all tests must pass (`go test ./...`)


mk-livepeer:
In an effort to move towards portability and continuous integration, we must make sure that all tests pass.

Developers should run:
```
go test ./...
```
before every check-in to ensure that all existing tests pass before proposing any pull requests.  Developers are encouraged to include tests to cover every testable change made to the repository.

For now, let's just make sure that the existing tests all pass.

Once this is done, we should add `go test ./...` to CI.

mk-livepeer:
Tests are still failing:
```
$ go test ./...
?   	github.com/livepeer/lpms/cmd/example	[no test files]
?   	github.com/livepeer/lpms/cmd/example/simple	[no test files]
?   	github.com/livepeer/lpms/cmd/transcoding	[no test files]
ok  	github.com/livepeer/lpms/core	0.043s
ok  	github.com/livepeer/lpms/ffmpeg	6.486s
HandShake: client signature does not match!
[hls @ 0xa00de00] Cannot use rename on non file protocol, this may lead to races and temporary partial files
WriteN, RTMP send error 32 (5 bytes)
WriteN, RTMP send error 32 (42 bytes)
RTMP_ReadPacket, failed to read RTMP packet body. len: 5604
HandShake: client signature does not match!
WriteN, RTMP send error 32 (5 bytes)
WriteN, RTMP send error 32 (42 bytes)
RTMP_ReadPacket, failed to read RTMP packet body. len: 5604
No application or playpath in URL!
RTMP_Connect0, failed to connect socket. 61 (Connection refused)
segmenter: Unable to open input
--- FAIL: TestNoRTMPListener (0.00s)
    video_segmenter_test.go:419: error was not nil; got  Unknown error occurred
HandShake: client signature does not match!
RTMP_ReadPacket, failed to read RTMP packet header
segmenter: Unable to open input
--- FAIL: TestServerDisconnect (0.11s)
    video_segmenter_test.go:457: Expected 'Input/output error' but instead got  Invalid data found when processing input
FAIL
FAIL	github.com/livepeer/lpms/segmenter	3.461s
ok  	github.com/livepeer/lpms/stream	0.241s
ok  	github.com/livepeer/lpms/transcoder	4.571s
ok  	github.com/livepeer/lpms/vidlistener	1.784s
ok  	github.com/livepeer/lpms/vidplayer	0.019s
```

j0sh:
[...]

PR #145, opened 2019-08-13
segmenter: Fix tests.

j0sh:
This feels a bit like a band-aid  -- if it's failing before writing the headers, is it really disconnecting "mid-stream" ? --  but seems to make the tests pass for now.

Fixes https://github.com/livepeer/lpms/issues/137
Fixes https://github.com/livepeer/lpms/issues/136

### DIFF
--- a/segmenter/video_segmenter_test.go
+++ b/segmenter/video_segmenter_test.go
@@ -425,18 +425,11 @@ type ServerDisconnectStream struct {
 }
 
 func (s *ServerDisconnectStream) ReadRTMPFromStream(ctx context.Context, dst av.MuxCloser) (chan struct{}, error) {
-	file, err := avutil.Open("test.flv")
-	if err != nil {
-		glog.Errorf("Error reading headers: %v", err)
-		return nil, err
-	}
-	header, err := file.Streams()
-	dst.WriteHeader(header)
 	dst.Close()
 	return make(chan struct{}), nil
 }
 
-func TestServerDisconnectMidStream(t *testing.T) {
+func TestServerDisconnect(t *testing.T) {
 	ffmpeg.InitFFmpeg()
 	port := "1938" // because we can't yet close the listener on 1935?
 	strm := &ServerDisconnectStream{}


segmenter/video_segmenter_test.go
```diff
@@ -431,8 +431,8 @@ func (s *ServerDisconnectStream) ReadRTMPFromStream(ctx context.Context, dst av.
 		return nil, err
 	}
 	header, err := file.Streams()
-	dst.WriteHeader(header)
 	dst.Close()
+	dst.WriteHeader(header)
```
darkdarkdragon: With this change this function as good as this:
```golang
func (s *ServerDisconnectStream) ReadRTMPFromStream(ctx context.Context, dst av.MuxCloser) (chan struct{}, error) {
	dst.Close()
	return make(chan struct{}), nil
}
```

Still, this doesn't test what it implies - it just closes connection before ffmpeg has chance to start processing actual packets 

> Curious, is the disconnect clean TCP/RTMP-wise?

For RTMP, I didn't found any 'closing' sequence in protocol description, so there is no 'unclean' disconnect RTMP-wise.

>  Is there a way we can hard-kill the server and have the client infer the disconnect, rather than having the connection tear itself down? That would capture the intent of this test better.

I did some research but wasn't able to find way to force TCP reset on connection from golang code.

j0sh: Good point, changed the function to just close immediately.

> Still, this doesn't test what it implies - it just closes connection before ffmpeg has chance to start processing actual packets

If we rename the test to remove the reference to "midstream" then that fits the fixed behavior of the test. tcpdump confirms the connection is established and we get the expected I/O error. How does that look?

j0sh:
Renamed the test itself to `TestServerDisconnect` so the scenario under test fits the description better.  It's not necessarily "mid-stream" but still tests a disconnect and results in an IO error as expected; the test itself doesn't change, just the name and how we're stubbing out the failures.

Rebased to master and addressed @darkdarkdragon comments. https://github.com/livepeer/lpms/commit/646b45763b0cabc77cdaa5d0aac33bc00467e817

Issue #125, opened 2019-05-30
Add helper to compute pixels in segment


yondonfu:
A broadcaster needs to be able to compute the pixels in a transcoded result for livepeer/go-livepeer#899.

This could involve adding a helper to the `ffmpeg` package:

```
func Pixels(input []byte) (int64, error)
```

This will likely require implementing the corresponding function in `ffmpeg/lpms_ffmpeg.c` that uses the ffmpeg C libraries for decoding the video and counting the frames.

j0sh:
Rather than having a helper specialized around pixels alone, the helper could compute and populate the fields that exist in the `TranscodeResult` struct, and the caller can extract the information it needs such as the pixel count.

j0sh:
Proposed implementation: to minimize additional C code [1], we can reuse our existing transcoder and pass in a null muxer along with codec copy [2] to minimize the work that needs to be done post-decode.

Building off the idea first proposed here https://github.com/livepeer/lpms/pull/146#pullrequestreview-275130706 : we can return a count of pixels decoded via the Transcode function, probably with the following signature:

```golang
type TranscodeResults {
  Decoded MediaInfo
  Encoded []MediaInfo
}

func Transcode4(input *TranscodeOptionsIn, ps []TranscodeOptions) (*TranscodeResults, error)
```

(May adjust the new Transcode3 function rather than introducing yet another one)

Working off https://github.com/livepeer/lpms/pull/122 - we can call the transcoder with codec copy and a null muxer. (ffmpeg has a null muxer, so we don't need to special-case any logic here.) Something like this:

```golang
type TranscodeOptions struct {
   // ... existing fields ...
   Encoder string
   Muxer string
}
func MediaInfo(string fname) (*MediaInfo, error) {
  in := &TranscodeOptionsIn{ Fname: fname }
  out := []TranscodeOptions{ TranscodeOptions { Oname: "-", Muxer: "null", Encoder: "copy" } }
  res, err := Transcode4(in, out)
  if err != nil {
    return err
  }
  return out[0].Decoded
}
```

[1] Minimizing the additional Cgo code not only improves composability and reusability, but also minimizes the error surface for accounting checks. Having separate implementations makes it more likely that counting discrepancies in one place or another would lead to accounting problems upstream. With this approach, any counting errors would be propagated through the system - so while we lose the double-checking that a separate implementation affords us, these errors should at least be consistent and not cause too many upstream discrepancies.

[...]

PR #146, opened 2019-08-14
Return count of encoded pixels from transcoder.

### LINKED ISSUE #124: Return pixels processed in FFMpegSegmentTranscoder.Transcode()

j0sh:
Fixes #124

### DIFF
--- a/cmd/transcoding/transcoding.go
+++ b/cmd/transcoding/transcoding.go
@@ -67,12 +67,15 @@ func main() {
 	ffmpeg.InitFFmpeg()
 
 	fmt.Printf("Setting fname %s encoding %d renditions with %v\n", fname, len(options), lbl)
-	err := ffmpeg.Transcode2(&ffmpeg.TranscodeOptionsIn{
+	res, err := ffmpeg.Transcode3(&ffmpeg.TranscodeOptionsIn{
 		Fname:  fname,
 		Accel:  accel,
 		Device: dev,
 	}, options)
 	if err != nil {
 		panic(err)
+	}
+	for i, r := range res {
+		fmt.Printf("profile=%v frames=%v pixels=%v\n", profiles[i].Name, r.Frames, r.Pixels)
 	}
 }
--- a/ffmpeg/ffmpeg.go
+++ b/ffmpeg/ffmpeg.go
@@ -41,6 +41,11 @@ type TranscodeOptions struct {
 	Device  string
 }
 
+type MediaInfo struct {
+	Frames int
+	Pixels int64
+}
+
 func RTMPToHLS(localRTMPUrl string, outM3U8 string, tmpl string, seglen_secs string, seg_start int) error {
 	inp := C.CString(localRTMPUrl)
 	outp := C.CString(outM3U8)
@@ -119,16 +124,20 @@ func accelDeviceType(accel Acceleration) (C.enum_AVHWDeviceType, error) {
 }
 
 func Transcode2(input *TranscodeOptionsIn, ps []TranscodeOptions) error {
+	_, err := Transcode3(input, ps)
+	return err
+}
 
+func Transcode3(input *TranscodeOptionsIn, ps []TranscodeOptions) ([]MediaInfo, error) {
 	if input == nil {
-		return ErrTranscoderInp
+		return nil, ErrTranscoderInp
 	}
 	if len(ps) <= 0 {
-		return nil
+		return nil, nil
 	}
 	hw_type, err := accelDeviceType(input.Accel)
 	if err != nil {
-		return err
+		return nil, err
 	}
 	fname := C.CString(input.Fname)
 	defer C.free(unsafe.Pointer(fname))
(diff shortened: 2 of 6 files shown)


ffmpeg/ffmpeg.go
```diff
@@ -41,6 +41,11 @@ type TranscodeOptions struct {
 	Device  string
 }
 
+type TranscodeResults struct {
```
j0sh: This name is what we agreed on in https://github.com/livepeer/lpms/issues/124 , but given https://github.com/livepeer/lpms/issues/125 I'm wondering if a more generic name like "MediaInfo" would be appropriate

ericxtang: I'm in favor of renaming to `MediaInfo`.

cmd/transcoding/transcoding.go
```diff
@@ -67,12 +67,15 @@ func main() {
...
-	err := ffmpeg.Transcode2(&ffmpeg.TranscodeOptionsIn{
+	res, err := ffmpeg.Transcode3(&ffmpeg.TranscodeOptionsIn{
 		Fname:  fname,
 		Accel:  accel,
 		Device: dev,
 	}, options)
 	if err != nil {
 		panic(err)
 	}
+	for i, r := range res {
```
mkrufky: should this logging always be enabled like this?   Can we put it behind a verbosity option?

ffmpeg/ffmpeg_test.go
```diff
@@ -374,3 +376,81 @@ func TestTranscoder_Timestamp(t *testing.T) {
...
+
+		// Check frame counts against ffprobe-reported output
+
+		// First, generate stats file
+		f, err := os.Create(fmt.Sprintf("%s/out%d.res.stats", dir, i))
+		if err != nil {
+			t.Error(err)
+		}
+		b := bufio.NewWriter(f)
+		fmt.Fprintf(b, `[STREAM]
```
mkrufky: style nit: please newline before the backtick if golang allows it.  (it might not, in which case, ignore this comment)

j0sh:
> every encode() call will be counting the pixels and frames regardless of whether we need it. @j0sh do you foresee this ever being a performance issue?

@ericxtang The overhead here is negligible  - it's just a bit of arithmetic -  so it won't be an issue, at least with just these fields. If we ever need to do some heavier computations, we can revisit whether to do this tracking unconditionally.

> when we want to count decoded pixels could we also populate the MediaInfo with the input pixels and frames by counting them in process_in()?

@yondonfu Yep - that part is easy enough. The rub here is what the `Transcode` API should look like. Currently, we return each rendition in its own `MediaInfo` struct:

```golang
 func Transcode3(input *TranscodeOptionsIn, ps []TranscodeOptions) ([]MediaInfo, error)
```

Where should we put the decoded pixel data?

* Within `MediaInfo` itself, something like a `decodedPixels`field . This feels sorta weird but would be "easy". Do we place this data in all renditions, or just one? 
* As an additional entry in the MediaInfo slice , perhaps at the beginning or the end. So for N renditions, we'd receive N+1 MediaInfo structs, with the additional one being the decoded information.
* Another wrapper struct. Something like this:

```golang
type TranscodeResults {
  Decoded MediaInfo
  Encoded []MediaInfo
}

func Transcode4(input *TranscodeOptionsIn, ps []TranscodeOptions) (*TranscodeResults, error)
```

It feels like we're getting a little heavy on the struct wrapping but that's probably the best method in this case.

> I agree with need to change name to MediaInfo, except that looks good.

@darkdarkdragon @ericxtang Changed 👍 

Also added a few more tests. Notably, there are scenarios where we'd encode *fewer* pixels than we originally requested due to maintaining the original aspect ratio during rescaling [1]. The implementation correctly counts the encoded pixels, but it seems like something that's easy to inadvertently break, so added a test for it.

As far as PM goes, this should only result in over-payment by overestimating the number of pixels; it shouldn't result in an under-payment. So if there's an aspect ratio mismatch between the source and transcoded segments, the B should end up with a larger credit after each segment for a given EV. cc @yondonfu 

[1] It'll scale to fit up to the desired resolution, but usually not beyond. The smaller side stays at whatever is proportionate for the original aspect ratio. Some odd adjustments could result in rounding up a pixel, but that shouldn't materially impact accounting.

cmd/transcoding/transcoding.go
j0sh: It's just a sample program to demonstrate the API, there aren't verbosity options

ffmpeg/ffmpeg_test.go
j0sh: After go fmt, the indentation is still strange:

```diff
[nix-shell:~/livepeer/code/github.com/livepeer/lpms/ffmpeg]$ git diff
diff --git a/ffmpeg/ffmpeg_test.go b/ffmpeg/ffmpeg_test.go
index 868eb20..a8756d2 100644
--- a/ffmpeg/ffmpeg_test.go
+++ b/ffmpeg/ffmpeg_test.go
@@ -435,7 +435,8 @@ func TestTranscoder_Statistics(t *testing.T) {
                        t.Error(err)
                }
                b := bufio.NewWriter(f)
-               fmt.Fprintf(b, `[STREAM]
+               fmt.Fprintf(b,
+                       `[STREAM]
 width=%d
 height=%d
 nb_read_frames=%d
```

I'm inclined to keep it as-is but can change if you prefer.

(The whole reason for the weird formatting here is because we test on `diff` output and golang doesn't have heredoc-style indentation adjustment. Originally I ignored whitespace via `diff -Wbu` but that felt somewhat fragile, so went this route.)

mkrufky: I think its better to go with go fmt, but I agree with you, it's weird either way.

VS code might try to autoformat it.  We should probably just commit whatever VSCode allows.  Do you use VSCode?

j0sh: vim for life :) `gofmt` is run after every save, so this is already "gofmt clean".

Does VSCode use a different formatter?

darkdarkdragon: No, VSCode uses same formatter. On re-saving this file it doesn't change formatting.

Issue #121, opened 2019-05-17
VAAPI Support


j0sh:
Supporting VAAPI accelerated transcoding would allow us to use AMD and  Intel QSV hardware. The support for VAAPI can mostly be laid on top of the current nvidia support. Some of the things that should be done:

* Update ffmpeg.go to add VAAPI accel type, and emit the `h264_vaapi` codec and `scale_vaapi` filters as needed

* Add PIX_FMT_VAAPI to init_video_filters

* Deal with the hwupload issue. For SW decode -> GPU rescale+encode, we avoid "wiring up" the GPU within LPMS by taking advantage of the `hwupload_cuda` filter which allows to specify a particular device to upload to:

https://github.com/livepeer/lpms/blob/77a56e05eba3017885e8a395136cf8726b0c65ee/ffmpeg/ffmpeg.go#L90-L94

However, the `device` option does not seem available within the plain `hwupload` filter that VAAPI uses. Our choices here would be to patch the `hwupload` filter to take a similar `device` option, or wire things together within LPMS (in a manner similar to how we wire in the decoder). The former seems much more preferable.

Note that SW decode -> GPU rescale+encode is a fringe case for us right now, but it works for nvidia and it would be nice to have symmetrical VAAPI features (otherwise it is sure to become tech debt).

* Check the logistics of building + running VAAPI-enabled binaries on non-VAAPI hardware. This works fine for nvidia, and I don't expect VAAPI to be much different, but it's something we should check. Also figure out whether it's better to depend on system shared libs (as we do for cuda) or statically link something like Mesa.

This is just from a quick skim of https://github.com/livepeer/lpms/pull/114/files ; probably missing steps and I'm certain things will break along the way.

wohlner:
This ticket is blocked because the office rig went offline when we moved NYC places.

PR #147, opened 2019-08-16
Count decoded pixels.

j0sh:
* Adds support for counting decoded pixels
* Allows passing in empty output parameters to force a decode-only flow.

The GetMediaInfo function in go-livepeer becomes:

```golang
func GetMediaInfo(fname string) (*MediaInfo, error) {
       in := &TranscodeOptionsIn{Fname: fname}
       res, err := Transcode3(in, nil)
       if err != nil {
               return nil, err
       }
       return &res.Decoded, nil
}
```

This could be further specialized to return pixels only, rather than a generic struct, because go-livepeer not likely to need the frame count.

Fixes https://github.com/livepeer/lpms/issues/125

### DIFF
--- a/cmd/transcoding/transcoding.go
+++ b/cmd/transcoding/transcoding.go
@@ -75,7 +75,8 @@ func main() {
 	if err != nil {
 		panic(err)
 	}
-	for i, r := range res {
+	fmt.Printf("profile=input frames=%v pixels=%v\n", res.Decoded.Frames, res.Decoded.Pixels)
+	for i, r := range res.Encoded {
 		fmt.Printf("profile=%v frames=%v pixels=%v\n", profiles[i].Name, r.Frames, r.Pixels)
 	}
 }
--- a/ffmpeg/ffmpeg.go
+++ b/ffmpeg/ffmpeg.go
@@ -46,6 +46,11 @@ type MediaInfo struct {
 	Pixels int64
 }
 
+type TranscodeResults struct {
+	Decoded MediaInfo
+	Encoded []MediaInfo
+}
+
 func RTMPToHLS(localRTMPUrl string, outM3U8 string, tmpl string, seglen_secs string, seg_start int) error {
 	inp := C.CString(localRTMPUrl)
 	outp := C.CString(outM3U8)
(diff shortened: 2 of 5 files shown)


j0sh:
Ready for review

darkdarkdragon:
@j0sh @yondonfu If I got it right, this will be called by broadcaster for every transcoded segment? So broadcaster will be doing software decoding of every transcoder stream? This will significantly increase CPU requirements, I'll need to re-test broadcaster's throughput with this turned on.

yondonfu:
@darkdarkdragon Yeah the current plan is for B to check the pixel count for each transcoded result. This will be async (see [this discussion thread](https://github.com/livepeer/go-livepeer/issues/899)). I'm interested in how this impacts the benchmarking results you've been doing and whether the performance hit is acceptable for now.

j0sh:
@darkdarkdragon Correct, in the short term, we will probably end up being CPU constrained rather than memory constrained on broadcast nodes.

It'll be interesting to see exactly how much this affects our capacity on hardware that would otherwise be memory-limited. This shouldn't be a blocker for anything right now though.

FWIW, it takes roughly 100-120ms to decode a 8-second video on my i7 Macbook Pro using software decoding. Cloud GPUs should be in that ballpark. Resolving the [LPMS bottleneck](https://github.com/livepeer/lpms/pull/141) should help decoder performance as well.

For the future we have a few possible ways to mitigate the performance impact here:

* Running the pixel counting on another machine (perhaps as a pre-check alongside verification)
* Running the decode on a Nvidia GPU if the broadcaster has one available
* Adding [VAAPI support](https://github.com/livepeer/lpms/issues/121) to the broadcaster to allow decode acceleration on Intel QuickSync or AMD VCE. (Or perhaps videotoolbox for MacOS specifically.) This wouldn't help cloud broadcast instances (which typically don't have these type of hardware functions exposed) but may help for local testing, or testing on dedicated hardware.

darkdarkdragon:
@j0sh as I understand, we need only number of frames? Can we parse video stream without decoding actual video data?

j0sh:
@darkdarkdragon We would need at least the resolution of each frame as well, since each frame could potentially have a different resolution. Rather than return a long list of frames, we just return a sum of pixels for the segment. This certainly isn't foolproof from a verification perspective, but the goals here are mostly around reconciliation for PM accounting, rather than verification.

It is possible that we parse the bitstream and extract the necessary information needed to determine the number of pixels, without going through a full decode (in H.264 this is mostly the macroblock count, eg `num_mbs_in_slice`). FFmpeg has an internal API for this but it isn't exposed for library consumers, and I'm not keen on us writing custom bitstream parsing code at the moment. For the time being, this was the quickest way to get the job done.

darkdarkdragon:
@j0sh thanks for the explanation! I didn't know that `each frame could potentially have a different resolution`. 

> FFmpeg has an internal API for this but it isn't exposed for library consumers,

Another reason to from FFmpeg, but certainly not now.

Issue #148, opened 2019-08-19
Support Stream Copy / Audio Verification


j0sh:
Audio is currently re-encoded as AAC for each segment in LPMS. Currently, we have no way to verify the audio track, and none is planned prior to the Streamflow release.

To guard against misbehavior around the audio stream, we should build stream copy support into LPMS. This will enable a few things:

* Audio copy. Then we can verify audio by running a bitexact comparison on the audio track only.
* The standalone Cgo segmenter can be subsumed by the main transcode function. This is a bonus, since stream copy is currently the only thing Cgo transcoder lacks to make it functionally equivalent to the segmenter.
* We can add a substrate for audio verification within LPMS if we add stream *dropping* support for the Transcode3 API.  The audio verification flow would consist of dropping unrelated streams (video), then hashing the remaining audio data, and comparing hashes against the source and encoded data. Roughly:

```
Encode:
  Encode video, copy audio

VerifyAudio:
  Copy audio, drop video
  Run through ffmpeg hash muxer

# For a given source segment, the B checks:
VerifyAudio(source) == VerifyAudio(Encode(source))
```

Most of the groundwork for this is complete (notably configurability for output muxers), although it is still missing the crucial stream copy portion: https://github.com/livepeer/lpms/commit/dc9f9e666b7d0199083ae6532bfd89534a33ff6c

Issue #55, opened 2018-02-15
Apply parameters via AVOptions


j0sh:
Rather than accessing the fields directly. Gives us much more flexibility in configuring the transcoder.

j0sh:
A brief explanation of how AVOptions works. Suppose we have this struct:
```
struct AVCodecContext {
   ..
   int bit_rate;
  ...
};
c = avcodec_alloc_context();
...
av_opt_set_int(c, "bit_rate", 100000, 0);
```

What `av_opt_set_int` translates to is essentially:

`c->bit_rate = 100000;`

Since AVOption accessors are already defined for all of the fields we'll need to set, it's much easier to use these when translating from our configuration struct, rather than enumerating each option ourselves. That way, users could also add in very particular codec settings that aren't available with one of our preset profiles. Naturally this will also require re-working how we do configuration.

Issue #54, opened 2018-02-14
Audio encoding parameters


j0sh:
We have several tunable parameters for video, but none for audio. We might want some, especially for downmixing. Eg, codec, sample rate, channels and channel layouts.

PR #150, opened 2019-08-29
Add stream copy and dropping.

j0sh:
Still needs checking for memory leaks, but opening this for review to get feedback.

This patch set modifies the `TranscodeOptions` struct to take options for the video encoder, audio encoder and muxer.  These options take a common `ComponentOpts` struct which specifies a name, and a set of options that correspond directly to the FFmpeg AVOptions for the component. Using these AVOptions should allow us to remove hardcoded and ad-hoc configuration options.  See "Follow Ups" for more details.

On top of the ComponentOpts, special-case handling is added for stream copy and stream dropping.

 This PR is backwards compatible with the existing Transcode3 API. If the audio/video options are omitted, the original behavior is preserved: x264 and AAC are encoded (or nvenc if hardware acceleration is specified).

### Examples

To copy an audio stream: `TranscodeOptions{ AudioEncoder: ComponentOptions{ Name: "copy" }}`

To discard a video stream: `TranscodeOptions{ VideoEncoder: ComponentOptions{ Name: "drop"}}`

To generate a md5 hash of the audio stream only:
```
TranscodeOptions{
  Oname: "audio.md5",
  AudioEncoder: ComponentOptions{ Name: "copy" },
  VideoEncoder: ComponentOptions{ Name: "drop" },
  Muxer: ComponentOptions{ Name: "md5" },
}
```

### Commits

https://github.com/livepeer/lpms/pull/150/commits/755f4b54c255379828f2213ab70753ef7aab5e1f : ffmpeg: Remove unused extradata copy code.

https://github.com/livepeer/lpms/pull/150/commits/a34627dfcddfabfd67e502ea0711c0442687682c : ffmpeg: Use AVOptions for muxer and A/V encoders.

https://github.com/livepeer/lpms/pull/150/commits/54db6b907b617d75e94c71e65f491b2284367a6f : ffmpeg: Add stream copy and drop.

### TODO before merging

- [x] Additional test cases as noted within code
- [ ] Valgrind/asan memory checking

### Fixes

Fixes https://github.com/livepeer/lpms/issues/148
Fixes https://github.com/livepeer/lpms/issues/55
Fixes https://github.com/livepeer/lpms/issues/54

### Follow Ups

Currently, the `Opts` part of the ComponentOptions is not used outside of test cases. Nearly all of our encoder configuration can be specified via AVOptions, however there is currently a dependency ordering issue: AVFilters are typically initialized prior to encoders, and the encoders take their configuration from the filters. Swapping the order of initialization should allow us to make use of AVOptions and minimize the use of ad-hoc, hard coded options as exhibited below:

https://github.com/livepeer/lpms/blob/f1b88b60503b33a74deebefadc7419fe8f49d146/ffmpeg/lpms_ffmpeg.c#L240-L244

https://github.com/livepeer/lpms/blob/f1b88b60503b33a74deebefadc7419fe8f49d146/ffmpeg/lpms_ffmpeg.c#L279-L283

https://github.com/livepeer/lpms/blob/f1b88b60503b33a74deebefadc7419fe8f49d146/ffmpeg/lpms_ffmpeg.h#L11-L17

### Notes

These ComponentOptions are passed in directly to FFmpeg, with all the flexibility and risk that entails. As such, library users (eg, go-livepeer) should take care to to ensure the options are used a safe manner, especially if exposed to arbitrary user input.

### DIFF
--- a/ffmpeg/ffmpeg.go
+++ b/ffmpeg/ffmpeg.go
@@ -28,6 +28,11 @@ const (
 	Amd
 )
 
+type ComponentOptions struct {
+	Name string
+	Opts map[string]string
+}
+
 type TranscodeOptionsIn struct {
 	Fname  string
 	Accel  Acceleration
@@ -39,6 +44,10 @@ type TranscodeOptions struct {
 	Profile VideoProfile
 	Accel   Acceleration
 	Device  string
+
+	Muxer        ComponentOptions
+	VideoEncoder ComponentOptions
+	AudioEncoder ComponentOptions
 }
 
 type MediaInfo struct {
@@ -87,6 +96,18 @@ func Transcode(input string, workDir string, ps []VideoProfile) error {
 		Accel: Software,
 	}
 	return Transcode2(inopts, opts)
+}
+
+func newAVOpts(opts map[string]string) *C.AVDictionary {
+	var dict *C.AVDictionary
+	for key, value := range opts {
+		k := C.CString(key)
+		v := C.CString(value)
+		defer C.free(unsafe.Pointer(k))
+		defer C.free(unsafe.Pointer(v))
+		C.av_dict_set(&dict, k, v, 0)
+	}
+	return dict
 }
 
 // return encoding specific options for the given accel
@@ -158,9 +179,12 @@ func Transcode3(input *TranscodeOptionsIn, ps []TranscodeOptions) (*TranscodeRes
 		if err != nil {
 			return nil, err
 		}
-		encoder, scale_filter, err := configAccel(input.Accel, p.Accel, input.Device, p.Device)
-		if err != nil {
-			return nil, err
+		encoder, scale_filter := p.VideoEncoder.Name, "scale"
+		if encoder == "" {
+			encoder, scale_filter, err = configAccel(input.Accel, p.Accel, input.Device, p.Device)
+			if err != nil {
+				return nil, err
+			}
 		}
 		// preserve aspect ratio along the larger dimension when rescaling
 		filters := fmt.Sprintf("fps=%d/%d,%s='w=if(gte(iw,ih),%d,-2):h=if(lt(iw,ih),%d,-2)'", param.Framerate, 1, scale_filter, w, h)
@@ -168,14 +192,33 @@ func Transcode3(input *TranscodeOptionsIn, ps []TranscodeOptions) (*TranscodeRes
 			// needed for hw dec -> hw rescale -> sw enc
 			filters = filters + ":format=yuv420p,hwdownload"
 		}
-		venc := C.CString(encoder)
+		muxOpts := C.component_opts{
+			opts: newAVOpts(p.Muxer.Opts), // don't free this bc of avformat_write_header API
+		}
+		if p.Muxer.Name != "" {
+			muxOpts.name = C.CString(p.Muxer.Name)
+			defer C.free(unsafe.Pointer(muxOpts.name))
+		}
+		vidOpts := C.component_opts{
+			name: C.CString(encoder),
+			opts: newAVOpts(p.VideoEncoder.Opts),
+		}
+		audioEncoder := p.AudioEncoder.Name
+		if audioEncoder == "" {
+			audioEncoder = "aac"
+		}
+		audioOpts := C.component_opts{
+			name: C.CString(audioEncoder),
+			opts: newAVOpts(p.AudioEncoder.Opts),
+		}
 		vfilt := C.CString(filters)
-		defer C.free(unsafe.Pointer(venc))
+		defer C.free(unsafe.Pointer(vidOpts.name))
+		defer C.free(unsafe.Pointer(audioOpts.name))
 		defer C.free(unsafe.Pointer(vfilt))
 		fps := C.AVRational{num: C.int(param.Framerate), den: 1}
 		params[i] = C.output_params{fname: oname, fps: fps,
 			w: C.int(w), h: C.int(h), bitrate: C.int(bitrate),
-			vencoder: venc, vfilters: vfilt}
+			muxer: muxOpts, audio: audioOpts, video: vidOpts, vfilters: vfilt}
 	}
 	var device *C.char
 	if input.Device != "" {
--- a/ffmpeg/ffmpeg.go
+++ b/ffmpeg/ffmpeg.go
@@ -172,12 +172,16 @@ func Transcode3(input *TranscodeOptionsIn, ps []TranscodeOptions) (*TranscodeRes
 		param := p.Profile
 		w, h, err := VideoProfileResolution(param)
 		if err != nil {
-			return nil, err
+			if "drop" != p.VideoEncoder.Name && "copy" != p.VideoEncoder.Name {
+				return nil, err
+			}
 		}
 		br := strings.Replace(param.Bitrate, "k", "000", 1)
 		bitrate, err := strconv.Atoi(br)
 		if err != nil {
-			return nil, err
+			if "drop" != p.VideoEncoder.Name && "copy" != p.VideoEncoder.Name {
+				return nil, err
+			}
 		}
 		encoder, scale_filter := p.VideoEncoder.Name, "scale"
 		if encoder == "" {
--- a/ffmpeg/ffmpeg_test.go
+++ b/ffmpeg/ffmpeg_test.go
@@ -575,3 +575,122 @@ func TestTranscoder_StatisticsAspectRatio(t *testing.T) {
 		t.Error(fmt.Errorf("Results did not match: %v ", r))
 	}
 }
+
+func TestTranscoder_MuxerOpts(t *testing.T) {
+	run, dir := setupTest(t)
+	defer os.RemoveAll(dir)
+
+	// Prepare test environment : truncate input file
+	cmd := `
+        set -eux
+        cd $0
+
+        cp "$1/../transcoder/test.ts" inp.ts
+        ffmpeg -i inp.ts -c:a copy -c:v copy -t 1 inp-short.ts
+    `
+	run(cmd)
+
+	prof := P240p30fps16x9
+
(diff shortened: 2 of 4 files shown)


ffmpeg/ffmpeg.go
```diff
@@ -28,6 +28,11 @@ const (
 	Amd
 )
 
+type ComponentOpts struct {
```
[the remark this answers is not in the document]
mkrufky: I might suggest `ComponentOptions` but I have no issue with the current name.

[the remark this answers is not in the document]
j0sh: Yes, that would be more consistent with our existing `TranscodeOptions` and `TranscodeOptionsIn` struct naming. Kinda long / wordy, but consistent.

Would really love to find a better word here than "component" though.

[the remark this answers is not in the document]
mkrufky: In M$ Windows BDA-land, we call them "filters" ...  The word "filter" might be a bit confusing to the open source world, however.  What do you think?   When you assemble the correct filters, the resulting pipeline is called a "graph".

Meanwhile, GSTREAMER is another framework that can be used for comparison.  In GSTREAMER, they're called "elements"

I see nothing wrong with the word "components" but maybe "elements" could be used if you're not happy with "components" ...  I am used to "filters" from my past of windows driver dev, but I don't think most users are familiar with the term.

ffmpeg/ffmpeg_test.go
```diff
@@ -575,3 +575,311 @@ func TestTranscoder_StatisticsAspectRatio(t *testing.T) {
 		t.Error(fmt.Errorf("Results did not match: %v ", r))
 	}
 }
+
```
mkrufky: I wonder if the things might be clearer during unit testing if this `ffmpeg_test.go` were broken down into smaller files.....

ffmpeg/ffmpeg.go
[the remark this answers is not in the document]
j0sh: Good to know what other projects call them! Agreed that using "filter" may be confusing, especially since we already use libavfilter here.

I'm leaning slightly into "components" over "elements" [1], but could go either way if you have a preference there. For now I've renamed things to "ComponentOptions", but that can be changed.

[1] The term "element" feels a bit more generic than "component", when it'd be good to have something more specific that describes their role in the system. Now that I think of it, it's kind of strange that there isn't an industry-standard term for these pieces of the transcoding pipeline!

ffmpeg/ffmpeg_test.go
j0sh: How should we split the file up? The segmenter- related tests could probably be put into its own file, but that's only ~100LOC or so. Pretty much everything else exercises various aspects of the transcoder. (That reminds me, these tests should probably be prefixed with `Transcoder_` to follow the naming convention in the rest of this file.)

There is some boilerplate here that I've been meaning to tidy up separately. Eg, basically all the shell tests start out with this line `set -eux ; cd $0` which can probably be extracted into a preamble that's automatically prefixed before `run`. Also the test file from `../transcoder/test.ts` is usually copied to the tmpdir and shortened to ~1 second; this can probably be done during the `setupTest` phase as well.

Using the testify package would probably save a few lines as well, but we don't really use that many assertions here.

j0sh:
> Is there a reason why TestStreamCopyAndDrop tests copy and drop together?

Yeah took a bit of a shortcut here. We need to verify stream copy works on audio and video separately, and it's easy to just drop the other stream via this new feature. So we kind of get a 2-for-1. Plus it saves some boilerplate, but that's not a great reason alone.

> I think it would be good to add individual unit tests for each as well

Sure - can add individual unit tests for drop and copy that don't assume the presence of the other feature.

ffmpeg/ffmpeg.go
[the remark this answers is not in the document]
mkrufky: I like components better than elements, but I don't feel strongly either way.

I think the industry standard is actually the Microsoft term, "filter" ...  However, M$ has become much less relevant in this space lately, and I don't think non-windows programmers like calling it a "filter".

Let's keep it as-is, "ComponentOptions" unless anybody feels differently

ffmpeg/ffmpeg_test.go
mkrufky: About the boilerplate stuff, it sounds like a good idea, and it should potentially make writing future tests easier.  Be sure to keep in mind that readability is the most important part of the tests - other developers will look at the unit tests in an effort to understand how each function works.  It sounds like extracting those few commands into a preamble would certainly help with this, but let's do it in a separate PR after this one is merged.

As for splitting the file up, it should also be handled separately after this PR is merged.  A nice thing about having many smaller test files rather than fewer larger test files is that when there are failures, they are easier to locate, due to the fact that the filename is displayed when a test fails.

100 LOC is not too small for a separate file...  If everything else is related to the transcoder, maybe there is another way to break it up.    If there isn't and we have to keep all the transcoder tests together, it's not the end of the world.... This was just a suggestion.

j0sh:
> Sure - can add individual unit tests for drop and copy that don't assume the presence of the other feature.

Turns out this exposed a failing case that wasn't covered by the other set of tests. Working to correct this.

ffmpeg/ffmpeg.go
```diff
@@ -151,31 +172,56 @@ func Transcode3(input *TranscodeOptionsIn, ps []TranscodeOptions) (*TranscodeRes
...
+		vidOpts := C.component_opts{
+			name: C.CString(encoder),
+			opts: newAVOpts(p.VideoEncoder.Opts),
+		}
+		audioEncoder := p.AudioEncoder.Name
+		if audioEncoder == "" {
+			audioEncoder = "aac"
+		}
+		audioOpts := C.component_opts{
+			name: C.CString(audioEncoder),
```
darkdarkdragon: Shouldn't this one be freed too?

ffmpeg/lpms_ffmpeg.c
```diff
@@ -248,49 +270,75 @@ static int open_output(struct output_ctx *octx, struct input_ctx *ictx)
...
-      // XXX only if transmuxing!
-      vc->extradata = av_mallocz(ictx->vc->extradata_size + AV_INPUT_BUFFER_PADDING_SIZE);
-      if (!vc->extradata) em_err("Unable to allocate video extradata\n");
-      memcpy(vc->extradata, ictx->vc->extradata, ictx->vc->extradata_size);
-      vc->extradata_size = ictx->vc->extradata_size;
-    }*/
-    ret = avcodec_open2(vc, codec, NULL);
+    ret = avcodec_open2(vc, codec, &octx->video->opts);
     if (ret < 0) em_err("Error opening video encoder\n");
+    av_dict_free(&octx->video->opts); // avcodec_open2 replaces this
```
darkdarkdragon: Wouldn't `octx->video->opts` be leaked in case of error?

ffmpeg/lpms_ffmpeg.c
```diff
@@ -248,49 +270,75 @@ static int open_output(struct output_ctx *octx, struct input_ctx *ictx)
...
     ac->sample_fmt = av_buffersink_get_format(octx->af.sink_ctx);
     ac->channel_layout = av_buffersink_get_channel_layout(octx->af.sink_ctx);
     ac->channels = av_buffersink_get_channels(octx->af.sink_ctx);
     ac->sample_rate = av_buffersink_get_sample_rate(octx->af.sink_ctx);
     ac->time_base = av_buffersink_get_time_base(octx->af.sink_ctx);
     if (fmt->flags & AVFMT_GLOBALHEADER) ac->flags |= AV_CODEC_FLAG_GLOBAL_HEADER;
-    ret = avcodec_open2(ac, codec, NULL);
+    ret = avcodec_open2(ac, codec, &octx->audio->opts);
     if (ret < 0) em_err("Error opening audio encoder\n");
+    av_dict_free(&octx->audio->opts); // avcodec_open2 replaces this
```
darkdarkdragon: Wouldn't `octx->audio->opts` be leaked in case of error?

ffmpeg/lpms_ffmpeg.c
```diff
@@ -248,49 +270,75 @@ static int open_output(struct output_ctx *octx, struct input_ctx *ictx)
...
+      // Sometimes the codec tag is wonky for some reason, so correct it
+      ret = av_codec_get_tag2(fmt->codec_tag, st->codecpar->codec_id, &st->codecpar->codec_tag);
+      //if (!ret) fprintf(stderr, "Audio codec tag not found. Continuing anyway\n");
+      avformat_transfer_internal_stream_timing_info(fmt, st, ist, AVFMT_TBCF_DEMUXER);
+    } else if (ac) {
+      st->time_base = ac->time_base;
+      ret = avcodec_parameters_from_context(st->codecpar, ac);
+      if (ret < 0) em_err("Error setting audio params from encoder\n");
+    } else em_err("No audio encoder; not a copy; what is this?\n");
     if (ret < 0) em_err("Unable to copy audio codec params\n");
```
darkdarkdragon: Looks like this will check `ret` from the call to `av_codec_get_tag2`?

ffmpeg/lpms_ffmpeg.c
```diff
@@ -303,8 +351,9 @@ static int open_output(struct output_ctx *octx, struct input_ctx *ictx)
     if (ret < 0) em_err("Error opening output file\n");
   }
 
-  ret = avformat_write_header(oc, NULL);
+  ret = avformat_write_header(oc, &octx->muxer->opts);
   if (ret < 0) em_err("Error writing header\n");
+  av_dict_free(&octx->muxer->opts); // avformat_write_header replaces this
```
darkdarkdragon: Wouldn't this will leak in case of error?

darkdarkdragon:
@LGTM
and we need to add version tag to this repository (to refer it from `go-livepeer`)

Issue #60, opened 2018-03-06
Handle video-only or audio-only streams in segmenter


j0sh:
Also consider cases where the video (or audio) is late; eg comes in past `analyzeduration` .

PR #151, opened 2019-09-05
WIP: Stream copy segmenter

j0sh:
Posting this PR here in case we need to pick it up someday.

This PR implements the segmenter by invoking the transcoder with stream copy and the corresponding HLS muxer options. Doing it this way allows a few things:

- Supports video-only or audio-only streams in segmenter. Fixes https://github.com/livepeer/lpms/issues/60

- Removes segmenter-specific FFmpeg code: `lpms_rtmp2hls` C function can be removed entirely, and `RTMPToHLS` [1] can be simplified to a single Transcode call.

[1] It should be noted that the `RTMPToHLS` function in particular is poorly named, since the input can come from anywhere, not just RTMP.

### TODO Before Merging

Some small changes are required to make the tests pass:

- [ ] Needs to drop late packets. Failing test `TestSegmenter_DropLatePackets`

https://github.com/livepeer/lpms/blob/f1b88b60503b33a74deebefadc7419fe8f49d146/ffmpeg/lpms_ffmpeg.c#L135

- [ ] Needs to drop leading frames up to a keyframe. Failing test `TestMissingKeyframe`

https://github.com/livepeer/lpms/blob/f1b88b60503b33a74deebefadc7419fe8f49d146/ffmpeg/lpms_ffmpeg.c#L130-L133

### DIFF
--- a/ffmpeg/ffmpeg.go
+++ b/ffmpeg/ffmpeg.go
@@ -61,6 +61,20 @@ type TranscodeResults struct {
 }
 
 func RTMPToHLS(localRTMPUrl string, outM3U8 string, tmpl string, seglen_secs string, seg_start int) error {
+	//glog.Errorf("Segmentating %s - %s - template %s", localRTMPUrl, outM3U8, tmpl)
+	in := &TranscodeOptionsIn{Fname: localRTMPUrl}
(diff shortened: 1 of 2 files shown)


criticaltv:
Audio only streams would be great to enable, as they offer a lower barrier to entry

Issue #152, opened 2019-09-05
Audio clicking on segmented transcodes


j0sh:
When transcoding audio with consecutive segments, harmonic content such as sine waves produces audible clicking when playing back transcoded results. This clicking shows up as a transient in spectrograms. Presumably this is because most audio encoders operate with a sliding window, which leads to discontinuities at the edges of each group of samples.

### Mitigation

One way to mitigate this effect is to pad the source segment with audio samples from adjacent segments. After transcoding, the padded samples can be dropped. However, it is possible that the efficiency of this depends on a few factors:

* The sample rate being constant between source and transcoded segments. If there is a change in sample rate (especially a downsample), it is possible there will be overlap with earlier segments.
* The required padding is understood. For AAC, 1024 samples at each end of the segment seems to be enough. However, may depend on the codec or the configuration of the codec. (Alternatively, three entire segments could be transcoded at once (prev/current/next) but this increases costs.)

This may have other implications in a live transcoding context, such as the need to wait for trailing padding to become available before submitting the source segment for transcoding.

Example of padding the end of a segment:
```bash
function pad {
  frames=1
  fname=$(basename $1)
  ffmpeg -loglevel warning -hide_banner -y -i $1 -muxdelay 0 -vn -c:a copy -copyts out/audio.ts
  ffmpeg -loglevel warning -hide_banner -y -i $2 -muxdelay 0 -vn -c:a copy -frames:a $frames -copyts out/end.ts
  cat out/audio.ts out/end.ts > out/padded_$fname
}

pad in/source_504.ts in/source_505.ts
pad in/source_505.ts in/source_506.ts
pad in/source_506.ts in/source_507.ts
pad in/source_507.ts in/source_508.ts
```

Example of trimming the padding:
```bash
function trim {
    fname=$(basename $1)
    frames=$(ffprobe -loglevel warning -hide_banner -count_packets \
      -select_streams a -show_streams $1 | grep nb_read_packets | grep -o '[0-9]*$')
    frames=$(($frames-1))
    ffmpeg -hide_banner -y -i $1 -muxdelay 0 -vn -c:a copy \
        -frames:a $frames -copyts out/stripped_$fname
}

trim out/padded_source_504.ts
trim out/padded_source_505.ts
trim out/padded_source_506.ts
trim out/padded_source_507.ts
```

### Samples

Source: `ffmpeg -f lavfi -i sine -c:a aac -f hls -hls_time 2 -t 10 test.m3u8`

[...]

PR #153, opened 2019-09-12
Fix frame dropping in filtergraph at EOF

j0sh:
ffmpeg: Fix frame dropping in filtergraph at EOF.
    
This is done by setting the EOF timestamp of the filtergraph
such that the timestamp is *after* the last frame inserted.
    
Also add tests exhibiting various cases uncovered in the process
of fixing this.

Fixes https://github.com/livepeer/lpms/issues/154

### DIFF
--- a/ffmpeg/ffmpeg_test.go
+++ b/ffmpeg/ffmpeg_test.go
@@ -24,6 +24,7 @@ func setupTest(t *testing.T) (func(cmd string), string) {
 	// The script is passed two arguments:
 	// a tempdir and the current working directory.
 	cmdFunc := func(cmd string) {
+		cmd = "cd $0 && set -eux;\n" + cmd
 		out, err := exec.Command("bash", "-c", cmd, dir, wd).CombinedOutput()
 		if err != nil {
 			t.Error(string(out[:]))
@@ -370,8 +371,8 @@ func TestTranscoder_Timestamp(t *testing.T) {
 		ffprobe -loglevel warning -select_streams v -show_streams -count_frames out0test.ts > test.out
 		grep avg_frame_rate=30 test.out
 		grep r_frame_rate=30 test.out
-		grep nb_read_frames=28 test.out
-		grep duration_ts=84000 test.out
+		grep nb_read_frames=29 test.out
+		grep duration_ts=87000 test.out
 		grep start_pts=138000 test.out
 	`
 	run(cmd)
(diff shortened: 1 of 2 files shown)


ffmpeg/ffmpeg_test.go
```diff
@@ -1037,3 +1048,292 @@ func TestTranscoder_StreamCopyAndDrop(t *testing.T) {
...
+	in.Fname = dir + "/out.mp4"
+	res2, err := Transcode3(in, nil)
+	if err != nil {
+		t.Error(err)
+	}
+	// TODO Ideally these two should match. As far as I can tell it is due
+	//      to timestamp rounding around EOF. Note this does not happen with
+	//      mpegts formatted output!
+	if res2.Decoded.Frames != 60 || res.Encoded[0].Frames != 61 {
+		t.Error("Did not get expected frame counts: is this fixed?!? ",
```
mkrufky: do we need the "is this fixed?!?" bit?

j0sh:
> I believe that the same affect achieved by this PR would be achieved by the pre-pending and post-pending of one audio frame

This PR doesn't attempt to address the audio popping issue; it only fixes frame dropping from the filtergraph as described in https://github.com/livepeer/lpms/issues/154 . The audio issue (along with your proposed fix) is still open and documented in https://github.com/livepeer/lpms/issues/152 .

ffmpeg/ffmpeg_test.go
j0sh: Happy to make adjustments to the language if you can think of a better phrasing :) Mostly, the intent was to express some notion of:

* This behavior is currently broken, and this unit test serves as reproducible documentation (along with https://github.com/livepeer/lpms/issues/155)
* But we need to make broken tests "pass"
* If the broken test no longer passes, maybe it means the underlying issue has been fixed? 🤞 
* If the test breaks, hopefully a distinct error message will prompt more careful review of the issue and any associated changes.

mkrufky: ok, that makes a lot more sense now - thank you for explaining :-)

darkdarkdragon:
@j0sh can't find way to comment on code that is not changed, so asking here:
here:
```c
  if (!filter || !filter->active) {
    // No filter in between decoder and encoder, so use input frame directly
    ret = encode(encoder, inf, octx, ost);
    if (ret < 0) return ret;
  }
```

shouldn't it exit unconditionally, like this:
```c
  if (!filter || !filter->active) {
    // No filter in between decoder and encoder, so use input frame directly
    ret = encode(encoder, inf, octx, ost);
    return ret;
  }
```

ffmpeg/ffmpeg_test.go
j0sh: Updated the language a bit in https://github.com/livepeer/lpms/pull/153/commits/ab98e5177bf328322e594a3efdc86a0dc3225ab2

j0sh:
> shouldn't it exit unconditionally, like this:

@darkdarkdragon Ah yes indeed! Fixed in https://github.com/livepeer/lpms/pull/153/commits/ab98e5177bf328322e594a3efdc86a0dc3225ab2

(Note that we don't actually hit this code path in LPMS itself but I've used it in the past for testing the C API directly...)

PR #159, opened 2019-10-31
Adding 25fps settings

ericxtang:


ericxtang:
Got ok from @j0sh in Discord

PR #160, opened 2019-11-21
Update joy4 and ffmpeg.

j0sh:
Splitting up some of https://github.com/livepeer/lpms/pull/141 into this PR in order to make the changes a touch less overwhelming.

* Updates FFmpeg to use the Livepeer forked branch at https://github.com/livepeer/ffmpeg
* Updates joy4 to fix some breakage resulting from the ffmpeg update. Specifically, this [change](http://git.videolan.org/?p=ffmpeg.git;a=commitdiff;h=ce1fcc8cede2971f6e36119e9dd41fb41b36c63a) modified the general libav code to *not* treat general IO errors as EOF. This broke some tests with the segmenter. Drilling down on the problem revealed that ffmpeg has a stricter view of what constitutes a proper EOF for RTMP : the playback stream needs to be torn down completely. Turns out that joy4 was simply terminating the connection, rather than tearing down the stream following the RTMP protocol. The teardown is added to joy4 here: https://github.com/livepeer/joy4/pull/6/commits/b2fea45cbded689f3235d48fdb28ce58518e6fce
* Fix some other test breakage induced by the ffmpeg update. In particular:

  * ADTS to ASC bitstream filters are now automatically inserted for the FLV muxer, so going from mpegts to FLV doesn't break anymore. Couldn't find another case which exhibits the same type of breakage, so just removed the test. (Almost everything else that holds AAC metadata in "Audio Specific Config" ASC format is a MP4 variant, and more exotic formats like nut can take basically anything.)

  * Some encoded file sizes got slightly larger. Haven't looked into precisely what / why but doesn't seem like an issue.

  * Tightened up the testing around draining GPU filter graphs. Something about the underlying implementation changed and is giving us 2 extra frames now. Sanity checked this against the results produced by the ffmpeg CLI itself: they match.

  * We have better test for RTMP stream disconnect. The old test broke since the stream is terminated cleanly now, even if no media packets are sent. We work around this by setting up a "server" with `nc` and sending a bogus RTMP handshake to which throws an IO error from ffmpeg. This isn't *strictly* a network disconnect, but is close enough.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ jobs:
   build:
     working_directory: /go/src/github.com/livepeer/lpms
     docker:
-      - image: "circleci/golang:1.11.5"
+      - image: "circleci/golang:1.13.4"
         environment:
           GOROOT: /usr/local/go
           PKG_CONFIG_PATH: "/home/circleci/compiled/lib/pkgconfig"
@@ -23,7 +23,7 @@ jobs:
           name: "Build FFMpeg"
           command: |
             sudo apt-get update
-            sudo apt-get install -y autoconf build-essential pkg-config autoconf gnutls-dev
+            sudo apt-get install -y autoconf build-essential pkg-config autoconf gnutls-dev zlib1g-dev netcat-openbsd
             bash ./install_ffmpeg.sh
 
       - save_cache:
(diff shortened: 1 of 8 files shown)


j0sh:
joy4 merged :+1:

PR #163, opened 2019-12-17
Preliminary Intel support

faeez-kadiri:


ffmpeg/ffmpeg.go
```diff
@@ -28,6 +28,7 @@ const (
 	Software Acceleration = iota
 	Nvidia
 	Amd
+	Intel
```
j0sh: Should we combine `Amd` and `Intel` into `VAAPI` ? Or will there be additional work required to support the AMD side of things?

ericxtang: @j0sh according to [this doc](https://trac.ffmpeg.org/wiki/HWAccelIntro), AMD seems to only get partial support in `VAAPI`.  So @faeez-kadiri and I had discussed to separate out the `Amd` implementation into a separate task for now to limit the scope of the current project.

j0sh: What is meant by partial support? What's missing? Note that page hasn't been updated in about a year [1], and [2] about two years.

The only reason I bring this up is because if we can keep the API surface area small by removing the AMD/Intel distinction then that's better overall. If using the ffmpeg CLI, there is no distinction between AMD/Intel, only the "VAAPI" hwaccel. 

For what it's worth, a GPU-only transcode test seemed to work just fine on AMD cards based on earlier testing using the ffmpeg CLI, so I don't see a reason why that wouldn't work here either.

[1] https://trac.ffmpeg.org/wiki/HWAccelIntro?action=history
[2] https://trac.ffmpeg.org/wiki/Hardware/VAAPI

PR #173, opened 2020-01-30
Close input in case of error

darkdarkdragon:
Check if buffer was actually allocated
In case `avformat_open_input` failed, close input that was
opened by `avio_open`
(`avformat_open_input` will not automatically close input
if it was provided to it)

### DIFF
--- a/ffmpeg/lpms_ffmpeg.c
+++ b/ffmpeg/lpms_ffmpeg.c
@@ -687,6 +687,7 @@ static int open_output(struct output_ctx *octx, struct input_ctx *ictx)
     if (av_buffersink_get_hw_frames_ctx(octx->vf.sink_ctx)) {
       vc->hw_frames_ctx =
         av_buffer_ref(av_buffersink_get_hw_frames_ctx(octx->vf.sink_ctx));
+      if (!vc->hw_frames_ctx) em_err("Unable to alloc hardware context\n");
     }
     vc->pix_fmt = av_buffersink_get_format(octx->vf.sink_ctx); // XXX select based on encoder + input support
     if (fmt->flags & AVFMT_GLOBALHEADER) vc->flags |= AV_CODEC_FLAG_GLOBAL_HEADER;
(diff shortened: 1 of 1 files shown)


ffmpeg/lpms_ffmpeg.c
```diff
@@ -834,7 +835,10 @@ static int open_input(input_params *params, struct input_ctx *ctx)
   ret = avio_open(&ic->pb, inp, AVIO_FLAG_READ);
   if (ret < 0) dd_err("demuxer: Unable to open file\n");
   ret = avformat_open_input(&ic, NULL, NULL, NULL);
-  if (ret < 0) dd_err("demuxer: Unable to open input\n");
+  if (ret < 0) {
+    avio_closep(&ic->pb); // need to close manually
```
j0sh: This will crash because `ic` is deallocated if `avformat_open_input` fails. In fact, CI is complaining about it.

Also, any error cleanup should go within `open_input_err`

darkdarkdragon: Fixed

ffmpeg/lpms_ffmpeg.c
```diff
@@ -847,6 +850,8 @@ static int open_input(input_params *params, struct input_ctx *ctx)
 
 open_input_err:
 fprintf(stderr, "Freeing input based on OPEN INPUT error\n");
+  avio_close(pb); // need to close manually, avformat_open_input
```
j0sh: Is the indentation here off?

j0sh: I see now that it's the fprintf which is indented wrong - feel free to fix that instead

darkdarkdragon: Fixed, waiting for the CI

j0sh: Also, just noticed that technically we could be passing in a null pointer here if `avformat_alloc_context` or `avio_open` fail, but it seems that `avio_close` already has a null check internally.

Defensively, I might do a `if (pb) avio_closep(&pb);` just to prevent that from ever blowing up in the future should anything change about avio's internal behavior, and to preemptively set `pb` to null. It's not strictly necessary here right now though; up to you.

darkdarkdragon: > Also, just noticed that technically we could be passing in a null pointer here if avformat_alloc_context or avio_open fail, but it seems that avio_close already has a null check internally.

I've checked this before committing - ffmpeg everywhere in the code relies on the ability to pass null pointer to `avid_close`. Say, `avformat_close_input` looks like this:
```c
void avformat_close_input(AVFormatContext **ps)
{
    AVFormatContext *s;
    AVIOContext *pb;

    if (!ps || !*ps)
        return;

    s  = *ps;
    pb = s->pb;

    if ((s->iformat && strcmp(s->iformat->name, "image2") && s->iformat->flags & AVFMT_NOFILE) ||
        (s->flags & AVFMT_FLAG_CUSTOM_IO))
        pb = NULL;

    flush_packet_queue(s);

    if (s->iformat)
        if (s->iformat->read_close)
            s->iformat->read_close(s);

    avformat_free_context(s);

    *ps = NULL;

    avio_close(pb);
}```

So I assumed that it is safe to pass null pointer there.

PR #176, opened 2020-03-20
WIP: Add MP4 format to VideoProfile.

j0sh:
-[ ] vidplayer changes still need tests


j0sh:
Added some more test cases in 7444a82 to ensure faststart is enabled (allows for playback of mp4s via http). Note this option generally involves a seek to the beginning of the file after the output has been written, which adds overhead on slow disks, and may be unsuitable if the output is being streamed over a non-seekable protocol.

j0sh:
Rebased with a number of changes in addition to what's already been mentioned:

* Renamed the `MP4` / `MPEGTS` enums to `FormatMP4` and `FormatMPEGTS` to avoid squatting on those generic names, which are not reuseable within the same package.

* Stricter default handling for unknown formats For example, in case we add a format to the enum and forget to add explicit handling in the transcoder for it.

* Added a lookup table going from format to extension (and corresponding inverse lookup)

* Use the lookup in `vidplayer` , and some cosmetic reindentation

darkdarkdragon:
Looks even better

Issue #129, opened 2019-07-01
Short Circuit Transcoding


j0sh:
Given a number of pixels as input, we should exit the transcode early if the output exceeds the pixels allowed. This allows the O to minimize the value at risk due to under-payment. 

Related: https://github.com/livepeer/lpms/issues/125 , https://github.com/livepeer/lpms/issues/124

PR #179, opened 2020-03-23
Persist filters in between segments.

j0sh:
This PR fixes GPU performance and VRAM growth issues by reusing the filtergraph in between segments. With this, LPMS is about equal to ffmpeg. Compared to previous LPMS versions, this gives us roughly a 30% performance gain for the workload of 1080p -> 720p,576p,460p,360p .

Note: Github for some reason refuses to show commits in the order they appear in the tree, and insists on incorrectly showing them chronologically based on commit time. The below is the correct ordering in tree, as will be reflected in a pull. The ordering is somewhat important, as some tests are added prior to the bulk of the functional modifications as a guard against behavioral changes. Ideally, nothing about the behavior as a result of this PR should change; it's just an extensive plumbing fix for performance.

Mostly ready for review, but not entirely comfortable with this yet ; in particular, SW encoding seems to show some differences at higher frame rates between 97c48b3237529e3e9e4ac6bd2ea217c8b415a1a4 and a03f838419bc0847c4922401d22f6b649883e1ff .

eab2b3705831fd205844598ba63bb387aeb51920 ffmpeg: Add tests for first and last PTS values. …
This helps ensure that changes to sensitive areas that munge pts
such as the filtergraph continue to produce the correct results

97c48b3237529e3e9e4ac6bd2ea217c8b415a1a4 ffmpeg: Check both software and Nvidia frame counts. …
More sanity checking.

a03f838419bc0847c4922401d22f6b649883e1ff ffmpeg: Persist filtergraph in between transcodes. …
Re-initializing the filtergraph turns out to be expensive,
especially if GPUs are in the picture. We get around 30% more
performance by persisting the filtergraph.

In order to flush the filter, we cache the most recent audio/video
frame and feed those into the filter repeatedly with a sentinel
value set (`AVFrame.opque`). Once we receive a frame from the
filtergraph with the sentinel set, we know the filter has been
completely flushed.

4ab522fd3346432dda5e4d752cdbba45aa113490 ffmpeg: Disallow out-of-order segments. …
This is a consequence of persisting the fps filter which expects
pts timestamps to be monotonic. Non-monotonic frames are dropped.

Accommodating out-of-order segments would involve totally
rewriting output timestamps based on the input value from the
source frame, while forging the pts that the filter receives.

00fe946e23c98458f39d6148c360f5c5ddfce716 ffmpeg: Log improvements and cosmetics.

00163c57d339ea9c0e23093bcbd1c64782475b97 ffmpeg: Documentation updates.


ffmpeg/lpms_ffmpeg.c
```diff
@@ -9,13 +9,52 @@
...
+//           For video, we cache the first packet we read (input_ctx.first_pkt).
+//           The pts is set to a sentinel value and fed to the decoder. Once we
+//           receive a frame from the decoder with the pts set to the sentinel
+//           value, then we know the decoder has been fully flushed. Ignore any
+//           subsequent frames we receive with the sentiel pts.
+//
+//  Filter:  The challenge here is around fps filter adding and dropping frames.
+//           The fps filter expects a strictly monotonic input pts: frames with
+//           earlier timestamps get dropped, and frames with too-late timestamps
+//           will see a bunch of duplicated frames be generated to catch up with
```
darkdarkdragon: What happens if segment missed? Like, orchestrator receives segment 1 and then segment 10?

ffmpeg/lpms_ffmpeg.c
```diff
@@ -1198,20 +1308,25 @@ int transcode(struct transcode_thread *h,
       // width / height will be zero for pure streamcopy (no decoding)
       decoded_results->frames += dframe->width && dframe->height;
       decoded_results->pixels += dframe->width * dframe->height;
+      has_frame = has_frame && dframe->width && dframe->height;
+      if (has_frame) last_frame = ictx->last_frame_v;
+    } else if (AVMEDIA_TYPE_AUDIO == ist->codecpar->codec_type) {
+      has_frame = has_frame && dframe->nb_samples;
+      if (has_frame) last_frame = ictx->last_frame_a;
+    }
       if (has_frame) {
```
darkdarkdragon: Can we fix indentation here?

ffmpeg/lpms_ffmpeg.c
j0sh: Great question. This actually works correctly. The fps filter duplicates the last frame to make up for the "missing frames" if it detects insertion(s) are necessary between the current frame and the last frame in order to maintain a smooth rate of output. If the current and last frames are "far apart", this does result in outputting a zillion duplicates to fill the gap.

But since the last frame is a flush frame, the duplicates are dropped before hitting the encoder. This leads to the final output still being correct. We actually have a test that implicitly checks this behavior via alternating timestamps in https://github.com/livepeer/lpms/pull/179/commits/4ab522fd3346432dda5e4d752cdbba45aa113490 which encodes segments 1 and 3 after dropping segments 0 and 2. But I'll see about adding an explicit check for skipped segments.

This brings up another issue I had overlooked until now. Since the fps filter comes *before* the scale filter, all duplicated frames get rescaled, then dropped. This is additional computation that really should be avoided since it is a clear attack vector: send artificial timestamps that are years apart and pin down the CPU / GPU as the filtergraph churns through all the duplicated frames.

https://github.com/livepeer/lpms/pull/179/commits/4ab522fd3346432dda5e4d752cdbba45aa113490

An easy fix for this would be to rearrange the filter string such that the fps adjustment comes *after* the rescaling, but it also means doing unnecessary work if the fps is being adjusted downwards (eg, 60fps -> 30fps) which is a rather common case (and that's the reason for the filters being ordered the way they are right now). But given the choice between doing redundant work, and protecting against an easy DoS,  it'd be better to rearrange the filters to avoid DoS.

That being said, I think it might be good to spend a little more time figuring out how to support alternating timestamps without making the code even more terrible than it already is, because I suspect we'll be running into ordering issues often with retries etc. Supporting alternating timestamps should allow us to preserve the current filtergraph ordering while mitigating the DoS.

Note that we've always had an issue with the user requesting an extremely high fps output, but theoretically that could be addressed by short circuiting the transcode as described in https://github.com/livepeer/lpms/issues/129

darkdarkdragon: > it'd be better to rearrange the filters to avoid DoS.

I agree we should do this until we implement support alternating timestamps.

darkdarkdragon: Also, this one https://github.com/livepeer/go-livepeer/issues/1422 will be an issue with persistent filtergraph too?

j0sh:
Rebased on top of master and squashed the fixups from earlier (and also probably some other small modifications which I've lost track of). But here are some relevant changes from the past day or so:

c5d18249a287a6ddb32c8d2630ea4b8c2507435e Filters rearranged due to the DoS issue mentioned in https://github.com/livepeer/lpms/pull/179#discussion_r401221104 and tests also added for skipped segments. The compute savings are real with the rearranged filters: with two skipped segments (240 frames), the test runs roughly 20% faster on my machine. Or seen in another light, we've avoided a 20% slowdown. But we may still end up a bit slower here due to redundant rescaling in downsampling the framerate.

6d66e5b293d98b2b56aa02a368758588db114103 Added a pts check for framerate pass through. Good thing this was added, because there was a bug in the timestamp rewriting logic.

6c4daa962c32ba624be04262d63db7cf8f4de0ee Aforementioned bug fixed. See the commit for details on what it does.

33c7a90debed18b0eaf6da6d02c9538d54961668 Cosmetic changes as requested in https://github.com/livepeer/lpms/pull/179#discussion_r400897919 .

> SW encoding seems to show some differences at higher frame rates between 97c48b3 and a03f838 .

This was a bit of a silly bug. The memory leak PR ( https://github.com/livepeer/lpms/pull/177/files ) unintentionally introduced a small change meant for this patch series (the `if(vf->active)` check), and things were working OK locally until I split the branches up (then they started failing in CI and locally). Merging #177 and rebasing this PR on top fixed the issue.

ffmpeg/lpms_ffmpeg.c
j0sh: Fixed in https://github.com/livepeer/lpms/commit/33c7a90debed18b0eaf6da6d02c9538d54961668

ffmpeg/lpms_ffmpeg.c
j0sh: > Also, this one livepeer/go-livepeer#1422 will be an issue with persistent filtergraph too?

It shouldn't, since timestamps are carried as int64 everywhere, and most sane formats can accommodate rollovers (including mpegts and RTMP). Dunno what the issue is with that ticket.

Filters rearranged and skip test added in https://github.com/livepeer/lpms/commit/c5d18249a287a6ddb32c8d2630ea4b8c2507435e

j0sh:
Rebased and merged.

PR #186, opened 2020-04-14
ffmpeg: Exit early for non-H264 input w/ NVDEC

### LINKED ISSUE #185: Processing non-supported codec leads to segfault with CUDA
To reproduce:
1. Generate video file `ffmpeg -loglevel quiet -f lavfi -i color=white:s=16x16 -vframes 1  -metadata title='!' -metadata service_provider='!' -f mpegts segment.ts`
2. Try to transcode using Nvidia acceleration.

Observe segment violation at https://github.com/FFmpeg/FFmpeg/blob/master/libavfilter/vf_scale_cuda.c#L259

This is because for `mpeg2video` codec our function `get_hw_pixfmt` not being called by `avcodec_open2` function, and so `hw_frames_ctx` field is left uninitialised.

jailuthra:
Resolves #185

Tested manually with the mpegts file in #185 - it now throws the correct error. (Still segfaults later though):
```$ ./transcoding segment.ts P240p25fps16x9 nv 0
Setting fname segment.ts encoding 1 renditions with nv
[mpegvideo @ 0x42b5480] Format mpegvideo detected only with low score of 12, misdetection possible!
Non H264 codec detected in input
Unable to open video decoder
Freeing input based on OPEN INPUT error
ERROR: logging before flag.Parse: E0414 18:54:09.197087  119394 ffmpeg.go:318] Transcoder Return : <nil>
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x5cdcae]

goroutine 1 [running]:
main.main()
            /home/darkapex/go/src/github.com/livepeer/lpms/cmd/transcoding/transcoding.go:78 +0x37e
```

### DIFF
--- a/ffmpeg/lpms_ffmpeg.c
+++ b/ffmpeg/lpms_ffmpeg.c
@@ -11,6 +11,7 @@
 
 // Not great to appropriate internal API like this...
 const int lpms_ERR_INPUT_PIXFMT = FFERRTAG('I','N','P','X');
+const int lpms_ERR_INPUT_CODEC = FFERRTAG('I','N','P','C');
 const int lpms_ERR_FILTERS = FFERRTAG('F','L','T','R');
 const int lpms_ERR_PACKET_ONLY = FFERRTAG('P','K','O','N');
 const int lpms_ERR_OUTPUTS = FFERRTAG('O','U','T','P');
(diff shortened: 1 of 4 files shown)


ffmpeg/lpms_ffmpeg.c
```diff
@@ -11,6 +11,7 @@
 
 // Not great to appropriate internal API like this...
 const int lpms_ERR_INPUT_PIXFMT = FFERRTAG('I','N','P','X');
+const int lpms_ERR_INPUT_CODEC = FFERRTAG('I','N','P','C');
```
j0sh, line 14: Usually for custom errors we need to add an error string for golang as well, see:

https://github.com/livepeer/lpms/blob/02ac425f2b6563a67189929a049dc489dbd936a9/ffmpeg/ffmpeg_errors.go#L31-L40

jailuthra:
Yep the missing error string was causing the latest segfault. Fixed it.

Also added a test case to verify the error thrown for unsupported codecs.

Issue #168, opened 2020-01-23
transcoding fails on segments containing very few video frames


darkdarkdragon:
Got one segment that transcodes ok with software transcoding, but transcoding on Nvidia card produces incorrect output
Ffprobe's output for transcoded segment:
```
[mpeg @ 0x561cfc3d6e00] Format mpeg detected only with low score of 25, misdetection possible!
[mp2 @ 0x561cfc3d8840] Header missing
    Last message repeated 2 times
[mpeg @ 0x561cfc3d6e00] decoding for stream 0 failed
[mpeg @ 0x561cfc3d6e00] Could not find codec parameters for stream 0 (Audio: mp2, 0 channels, s16p): unspecified frame size
Consider increasing the value for the 'analyzeduration' and 'probesize' options
Input #0, mpeg, from 'out_nv_0_out.ts':
  Duration: 00:00:00.02, start: 634.176011, bitrate: 453 kb/s
    Stream #0:0[0x1c0]: Audio: mp2, 0 channels, s16p
```
source segment:
[bad_video_bbb_sunflower_1080p_30fps_normal_t02mp4_057c3e76c81edee3a550_0_0_317_source.ts.gz](https://github.com/livepeer/lpms/files/4105368/bad_video_bbb_sunflower_1080p_30fps_normal_t02mp4_057c3e76c81edee3a550_0_0_317_source.ts.gz)
transcoded on Nvidia:
[out_nv_0_out.ts.gz](https://github.com/livepeer/lpms/files/4105375/out_nv_0_out.ts.gz)
software transcoded:
[out_sw_0_out.ts.gz](https://github.com/livepeer/lpms/files/4105376/out_sw_0_out.ts.gz)

j0sh:
Nice one, this looks like a bug. There are only 3 video frames in the source, and the decoder doesn't return any of them by the time reading completes. Looks like we aren't flushing where we should in this case, because with nvidia, we depend on getting that first frame to use as a sentinel in order to drain the decoder.

Will have to think about this one. We might not necessarily want to close the decoder outright to flush, since re-opening it is expensive.

darkdarkdragon:
[bad_video_ccb0eebc-2f55-497c-b10c-3a02cd578ba9_317_source.ts.gz](https://github.com/livepeer/lpms/files/4188982/bad_video_ccb0eebc-2f55-497c-b10c-3a02cd578ba9_317_source.ts.gz)

[...]

PR #187, opened 2020-04-21
ffmpeg: use NULL flushing for decoder

jailuthra:
**WIP**

Passing the framecount to `process_in` feels unclean.

Should we instead keep a flag called `needs_null_flushing` within input_ctx, and let process_in set it to false once it sees that CUDA has started returning frames?

This can also replace the new enum I think.

### DIFF
--- a/ffmpeg/lpms_ffmpeg.c
+++ b/ffmpeg/lpms_ffmpeg.c
@@ -60,6 +60,13 @@ const int lpms_ERR_DTS = FFERRTAG('-','D','T','S');
 //
 // Internal transcoder data structures
 //
+
+enum LPMSFlushMode {
+  LPMS_FLUSH_UNDEFINED = 0,
+  LPMS_FLUSH_SENTINEL,
+  LPMS_FLUSH_DEFAULT
+};
+
 struct input_ctx {
   AVFormatContext *ic; // demuxer required
   AVCodecContext  *vc; // video decoder optional
@@ -75,6 +82,7 @@ struct input_ctx {
   // Decoder flush
   AVPacket *first_pkt;
   int flushed;
+  enum LPMSFlushMode flush_mode;
 
   // Filter flush
   AVFrame *last_frame_v, *last_frame_a;
@@ -955,7 +963,7 @@ open_input_err:
 #undef dd_err
 }
 
-int process_in(struct input_ctx *ictx, AVFrame *frame, AVPacket *pkt)
+int process_in(struct input_ctx *ictx, int framecount, AVFrame *frame, AVPacket *pkt)
 {
 #define dec_err(msg) { \
   if (!ret) ret = -1; \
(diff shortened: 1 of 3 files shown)


j0sh:
Would it be simpler to call `process_in` as needed with the `pkt` parameter set to NULL?

Note that the `while(1)` within `process_in` isn't needed anymore; so things could be changed to something like:

```
if (!pkt) {
  ret = avcodec_send_packet(decoder, NULL);
  if (ret < 0) dd_err("Error flushing decoder\n");
} else {
  av_init_packet(pkt);
  // existing code within while(1)
}
```

Then I think the rest of the code should still work unchanged from there, and the decision on whether to begin flushing can be done from outside, eg within `transcode` rather than `process_in` itself.

That aside, was there any luck with using `avcodec_flush_buffers` in place of sending a NULL packet ?

jailuthra:
> Then I think the rest of the code should still work unchanged from there, and the decision on whether to begin flushing can be done from outside, eg within `transcode` rather than `process_in` itself.

This sounds neat. I'll split a function called `flush_buffers` out from `process_in` and let the caller decide what to trigger. Flushing logic is anyway complicated if we include the sentinel method, so it would fit neatly outside.

> That aside, was there any luck with using `avcodec_flush_buffers` in place of sending a NULL packet ?

I think there's a misunderstanding. What I understood from [[1]](https://ffmpeg.org/doxygen/trunk/group__lavc__encdec.html) is we still need to send NULL packets to flush the decoder buffers. Once they're all  flushed we can resume normal decoding using `avcodec_flush_buffers()`

I'm still trying it out... Need to check when flush is complete (i.e. we receive EOF instead of EAGAIN)

j0sh:
> I'll split a function called flush_buffers out from process_in and let the caller decide what to trigger

Sure - just keep in mind that first we should attempt the smallest change necessary to fix the issue. Then afterwards we can look at extracting the other parts from `process_in`, eg sentinel frame handling. (I suspect the proposed fix should not really affect how sentinel frames currently work, but might be wrong.)

> What I understood from [1] is we still need to send NULL packets to flush the decoder buffers. Once they're all flushed we can resume normal decoding using avcodec_flush_buffers()

Ah I see, got it 👍

jailuthra:
After latest commit in this PR, it passes all tests. Mitigated the issue with segment order checking code.

Current state of PR:
* Uses NULL flushing always for SW decoder, and for HW decoder as well if it is still buffering frames, and didn't start outputting.
* Passes all tests (like re-using transcoder for multiple segments).
* Mitigated the issue with segment order checking code in previous commit (see description for more)

j0sh:
Looks good - needs unit tests though, and I think we have a lot of scenarios to check. Here's a start - https://gist.github.com/j0sh/02daa8928e80f5a35e23499b9c54c590

(I suspect your fix will mitigate the failures I'm seeing in my tests.)

ffmpeg/lpms_ffmpeg.c
```diff
@@ -1335,6 +1357,7 @@ int transcode(struct transcode_thread *h,
     has_frame = lpms_ERR_PACKET_ONLY != ret;
 
     if (AVMEDIA_TYPE_VIDEO == ist->codecpar->codec_type) {
+      if (ictx->flushing) dframe->pkt_dts = ictx->first_pkt->dts;
```
j0sh, line 1360: Curious - what does this fix? Maybe good to note this in a code comment.

ffmpeg/lpms_ffmpeg.c
```diff
@@ -1021,18 +1030,31 @@ int process_in(struct input_ctx *ictx, AVFrame *frame, AVPacket *pkt)
...
+      // To accommodate CUDA, we feed the decoder a a sentinel (flush) frame.
+      // Once the flush frame has been decoded, the decoder is fully flushed.
+      send_first_pkt(ictx);
+
+      ret = avcodec_receive_frame(ictx->vc, frame);
+      if (!ret && is_flush_frame(frame))  ictx->flushed = 1;
+    } else {
+      // Classic flushing by sending NULL packets to decoder
+      ictx->flushing = 1;
+      ret = avcodec_send_packet(ictx->vc, NULL);
```
j0sh, line 1048: Should `ret` be checked for errors here? If it doesn't need to be checked, maybe we can just not set it?

ffmpeg/lpms_ffmpeg.c
jailuthra: This is used to keep the [segment-ordering check](https://github.com/livepeer/lpms/blob/6e167971ec68573d2cdf47cbb0b163fbb3a981a7/ffmpeg/lpms_ffmpeg.c#L995) work same as it did before.

FFmpeg sets pkt_dts to the dts of the [packet triggering the frame output](https://ffmpeg.org/doxygen/trunk/structAVFrame.html#aa52951f35ec9e303d3dfeb4b3e44248a) and *not* the packet that was actually decoded to get the frame.

If we don't set it to first_pkt->dts when using NULL-flushing, the last_frame->pkt_dts is set to 0 for the flushed frames and the segment-ordering check fails.

I tried modifying the segment-ordering check but that caused other tests to fail, and fixing those triggered more. So for this PR I'll avoid going into that rabbit hole.

I'll add a comment for this as well.

ffmpeg/lpms_ffmpeg.c
jailuthra: ah yes it is not needed, i had it for debugging then forgot.

jailuthra:
Latest State of PR:
* Use NULL flushing for SW/NV decoder all the time
* Mitigate 1-frame segment drops during fps filtering by using output duration in the sentinel while flushing
* Add unit tests for short segments (with parameterized frame count and accel mode)
* Passes all tests. Fixes #168

Will rename/squash commits if they're ready to ship.

jailuthra:
Closing, as #189 fixed the original issue

Issue #155, opened 2019-09-17
Number of encoded frames does not match number of decoded frames


j0sh:
Under particular conditions, the number of encoded frames does not match number of decoded frames.

* Reproducible case that encodes 61 "frames" (61 frames encoded and 61 packets muxed) but only decodes 60, as shown below.
* FFProbe  mirrors our numbers: it reads the same 61 packets in the bitstream, but only decodes 60 frames.
* This only seems to happen with the MP4 muxer. The mpegts muxer works as expected.
* Possible that this is a separate slice / field / etc but this doesn't seem likely.
* If this is indeed a non-display frame of some sort : is there a way to detect this?
  https://github.com/livepeer/lpms/blob/d722afc1dfa0e7b7102db8b67588a00be6dfd938/ffmpeg/ffmpeg_test.go#L1117-L1209

PR #189, opened 2020-05-06
ffmpeg: Fix short segments flushing without using NULL packets

jailuthra:
* Maintain diff of sent packets/recv frames to know when decoder is done flushing.
* Stop sending sentinel packets after n=5 attempts of not getting any valid frame back.
* Fix fps filter dropping 1-frame segments by sending a flush frame with a longer pts.
* Add parameterized unit tests to reproduce #168 

In #187 I tried to fix the short segment flushing using NULL packets, but it had heavy performance impact. So I've created this fresh PR because that got very messy.

I tried my best to benchmark this against master and it seems to perform equally well. I noob though, pls verify :)

### DIFF
--- a/ffmpeg/api_test.go
+++ b/ffmpeg/api_test.go
@@ -478,3 +478,206 @@ func TestTranscoder_API_AlternatingTimestamps(t *testing.T) {
   `
 	run(cmd)
 }
+
+// test short segments
+func shortSegments(t *testing.T, accel Acceleration, fc int) {
+	run, dir := setupTest(t)
+	defer os.RemoveAll(dir)
+
+	cmd := `
+    # generate segments with #fc frames
+    cp "$1/../transcoder/test.ts" .
+	frame_count=%d
+
+    ffmpeg -loglevel warning -ss 0 -i test.ts -c copy -frames:v $frame_count -copyts short0.ts
+    ffmpeg -loglevel warning -ss 2 -i test.ts -c copy -frames:v $frame_count -copyts short1.ts
+    ffmpeg -loglevel warning -ss 4 -i test.ts -c copy -frames:v $frame_count -copyts short2.ts
+    ffmpeg -loglevel warning -ss 6 -i test.ts -c copy -frames:v $frame_count -copyts short3.ts
+
+    ffprobe -loglevel warning -count_frames -show_streams -select_streams v short0.ts | grep nb_read_frames=$frame_count
+    ffprobe -loglevel warning -count_frames -show_streams -select_streams v short1.ts | grep nb_read_frames=$frame_count
+    ffprobe -loglevel warning -count_frames -show_streams -select_streams v short2.ts | grep nb_read_frames=$frame_count
+    ffprobe -loglevel warning -count_frames -show_streams -select_streams v short3.ts | grep nb_read_frames=$frame_count
+  `
+	run(fmt.Sprintf(cmd, fc))
+
+	// Test if decoding/encoding expected number of frames
+	tc := NewTranscoder()
+	defer tc.StopTranscoder()
+	for i := 0; i < 4; i++ {
+		fname := fmt.Sprintf("%s/short%d.ts", dir, i)
+		t.Log("fname ", fname)
+		in := &TranscodeOptionsIn{Fname: fname, Accel: accel}
+		out := []TranscodeOptions{{Oname: dir + "/out.ts", Profile: P144p30fps16x9, Accel: accel}}
+		res, err := tc.Transcode(in, out)
+		if err != nil {
+			t.Error(err)
+		}
+		if fc != res.Decoded.Frames {
+			t.Error("Did not decode expected number of frames: ", res.Decoded.Frames)
+		}
+		if 0 == res.Encoded[0].Frames {
+			// TODO not sure what should be a reasonable number here
+			t.Error("Did not encode any frames: ", res.Encoded[0].Frames)
+		}
+	}
+
+	// test standalone stream copy
+	tc.StopTranscoder()
+	tc = NewTranscoder()
+	for i := 0; i < 4; i++ {
+		fname := fmt.Sprintf("%s/short%d.ts", dir, i)
+		oname := fmt.Sprintf("%s/vcopy%d.ts", dir, i)
+		in := &TranscodeOptionsIn{Fname: fname, Accel: accel}
+		out := []TranscodeOptions{
+			{
+				Oname: oname,
+				VideoEncoder: ComponentOptions{Name: "copy", Opts: map[string]string{
+					"mpegts_flags": "resend_headers,initial_discontinuity",
+				}},
+				Accel: accel,
+			},
+		}
+		res, err := tc.Transcode(in, out)
+		if err != nil {
+			t.Error(err)
+		}
+		if res.Encoded[0].Frames != 0 {
+			t.Error("Unexpected frame counts from stream copy")
+			t.Error(res)
+		}
+		cmd = `
+        # extract video track, compare md5sums
+		i=%d
+        ffmpeg -i short$i.ts -an -c:v copy -f md5 short$i.md5
+        ffmpeg -i vcopy$i.ts -an -c:v copy -f md5 vcopy$i.md5
+        diff -u short$i.md5 vcopy$i.md5
+        `
+		run(fmt.Sprintf(cmd, i))
+	}
+
+	// test standalone stream drop
+	tc.StopTranscoder()
+	tc = NewTranscoder()
+	for i := 0; i < 4; i++ {
+		fname := fmt.Sprintf("%s/short%d.ts", dir, i)
+		oname := fmt.Sprintf("%s/vdrop%d.ts", dir, i)
+		// Normal case : drop only video
+		in := &TranscodeOptionsIn{Fname: fname, Accel: accel}
+		out := []TranscodeOptions{
+			{
+				Oname:        oname,
+				VideoEncoder: ComponentOptions{Name: "drop"},
+				Accel:        accel,
+			},
+		}
+		res, err := tc.Transcode(in, out)
+		if err != nil {
+			t.Error(err)
+		}
+		if res.Decoded.Frames != 0 || res.Encoded[0].Frames != 0 {
+			t.Error("Unexpected count of decoded frames ", res.Decoded.Frames, res.Decoded.Pixels)
+		}
+
+	}
+
+	// test framerate passthrough
+	tc.StopTranscoder()
+	tc = NewTranscoder()
+	for i := 0; i < 4; i++ {
+		fname := fmt.Sprintf("%s/short%d.ts", dir, i)
+		oname := fmt.Sprintf("%s/vpassthru%d.ts", dir, i)
+		out := []TranscodeOptions{{Profile: P144p30fps16x9, Accel: accel}}
+		out[0].Profile.Framerate = 0 // Passthrough!
+
+		out[0].Oname = oname
+		in := &TranscodeOptionsIn{Fname: fname, Accel: accel}
+		res, err := tc.Transcode(in, out)
+		if err != nil {
+			t.Error("Could not transcode: ", err)
+		}
+		// verify that output frame count is same as input frame count
+		if res.Decoded.Frames != fc || res.Encoded[0].Frames != fc {
+			t.Error("Did not get expected frame count; got ", res.Encoded[0].Frames)
+		}
+	}
+
+	// test low fps (3) to low fps (1)
+	tc.StopTranscoder()
+	tc = NewTranscoder()
+
+	cmd = `
+	frame_count=%d
+	# convert segment to 3fps and trim it to #fc frames
+	ffmpeg -loglevel warning -i test.ts -vf fps=3/1 -c:v libx264 -c:a copy -frames:v $frame_count short3fps.ts
+
+	# sanity check
+	ffprobe -loglevel warning -show_streams short3fps.ts | grep r_frame_rate=3/1
+	ffprobe -loglevel warning -count_frames -show_streams -select_streams v short3fps.ts | grep nb_read_frames=$frame_count
+  `
+	run(fmt.Sprintf(cmd, fc))
+
+	fname := fmt.Sprintf("%s/short3fps.ts", dir)
+	in := &TranscodeOptionsIn{Fname: fname, Accel: accel}
+	out := []TranscodeOptions{{Oname: dir + "/out1fps.ts", Profile: P144p30fps16x9, Accel: accel}}
+	out[0].Profile.Framerate = 1 // Force 1fps
+	res, err := tc.Transcode(in, out)
+	if err != nil {
+		t.Error(err)
+	}
+	if fc != res.Decoded.Frames {
+		t.Error("Did not decode expected number of frames: ", res.Decoded.Frames)
+	}
+	if 0 == res.Encoded[0].Frames {
+		t.Error("Did not encode any frames: ", res.Encoded[0].Frames)
+	}
+
+	// test a bunch of weird cases together
+	tc.StopTranscoder()
+	tc = NewTranscoder()
+	profile_low_fps := P144p30fps16x9
+	profile_low_fps.Framerate = uint(fc) // use the input frame count as the output fps, why not
+	profile_passthrough_fps := P144p30fps16x9
+	profile_passthrough_fps.Framerate = 0
+	for i := 0; i < 4; i++ {
+		fname := fmt.Sprintf("%s/short%d.ts", dir, i)
+		in := &TranscodeOptionsIn{Fname: fname, Accel: accel}
+		out := []TranscodeOptions{
+			{
+				Oname:        fmt.Sprintf("%s/lowfps%d.ts", dir, i),
+				Profile:      profile_low_fps,
+				AudioEncoder: ComponentOptions{Name: "copy"},
+				Accel:        accel,
+			},
+			{
+				Oname:        fmt.Sprintf("%s/copyall%d.ts", dir, i),
+				VideoEncoder: ComponentOptions{Name: "copy"},
+				AudioEncoder: ComponentOptions{Name: "drop"},
+				Accel:        accel,
+			},
+			{
+				Oname:   fmt.Sprintf("%s/passthru%d.ts", dir, i),
+				Profile: profile_passthrough_fps,
+				Accel:   accel,
+			},
+		}
+		res, err := tc.Transcode(in, out)
+		if err != nil {
+			t.Error(err)
+		}
+		if res.Encoded[0].Frames == 0 || res.Encoded[1].Frames != 0 || res.Encoded[2].Frames != fc {
+			t.Error("Unexpected frame counts from short segment copy-drop-passthrough case")
+			t.Error(res)
+		}
+	}
+
+}
+
+func TestTranscoder_ShortSegments(t *testing.T) {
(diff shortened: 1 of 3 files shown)


ffmpeg/ffmpeg_test.go
```diff
@@ -1512,3 +1512,60 @@ func TestTranscoder_FormatOptions(t *testing.T) {
...
+		in := &TranscodeOptionsIn{Fname: fname, Accel: accel}
+		out := []TranscodeOptions{{Oname: dir + "/out.ts", Profile: P144p30fps16x9, Accel: accel}}
+		res, err := tc.Transcode(in, out)
+		if err != nil {
+			t.Error(err)
+		}
+		if fc != res.Decoded.Frames {
+			t.Error("Did not decode expected number of frames: ", res.Decoded.Frames)
+		}
+		if 0 == res.Encoded[0].Frames {
```
j0sh, line 1552: Is there a predictable pattern to the number of encoded frames that we can check against here?

ffmpeg/ffmpeg_test.go
```diff
@@ -1512,3 +1512,60 @@ func TestTranscoder_FormatOptions(t *testing.T) {
 		t.Error("Did not get expected error with invalid format ", err)
 	}
 }
+
+// test short segments
+func shortSegments(t *testing.T, accel Acceleration, fc int) {
```
j0sh, line 1517: I've been putting tests using the "new API" (NewTranscoder / Transcode / StopTranscoder) under `api_test.go` partially because this file was getting large. But up to you.

ffmpeg/lpms_ffmpeg.c
```diff
@@ -75,6 +75,16 @@ struct input_ctx {
...
   AVPacket *first_pkt;
   int flushed;
+  int flushing;
+  // The diff of `packets sent - frames recv` serves as an estimate of
+  // internally buffered packets by the decoder.  We're done flushing when this
+  // becomes 0.
+  uint16_t pkt_diff;
+  // We maintain a count of sentinel packets sent without receiving any
+  // valid frames back, and stop flushing if it crosses SENTINEL_MAX.
+  // FIXME This is needed due to issue #155 - input/output frame mismatch.
```
j0sh, line 85: Do you think that once #155 is fixed, we won't need to track `sentinel_count` anymore?

ffmpeg/lpms_ffmpeg.c
```diff
@@ -1148,11 +1177,17 @@ int process_out(struct input_ctx *ictx, struct output_ctx *octx, AVCodecContext
...
-    filter->flush_offset += frame->pkt_duration;
+    // Increment pts offset by:
+    // - input packet's duration usually
+    int64_t dur = frame->pkt_duration;
+    // - output packet's duration if using fps filter & haven't encoded anything yet
+    if (is_video && octx->fps.den && !octx->res->frames) {
+      AVStream *ist = ictx->ic->streams[ictx->vi];
+      dur = av_rescale_q(frame->pkt_duration, ist->r_frame_rate, octx->fps);
+    }
+    filter->flush_offset += dur;
```
j0sh, line 1190: Here we increment `filter->flush_count += frame->pkt_duration` ; do we have to make a similar adjustment there as well?

https://github.com/livepeer/lpms/blob/42a805ec2debbfa6b448a28b16650f4e186dcb5e/ffmpeg/lpms_ffmpeg.c#L1229

ffmpeg/ffmpeg_test.go
```diff
@@ -1512,3 +1512,60 @@ func TestTranscoder_FormatOptions(t *testing.T) {
...
+		if fc != res.Decoded.Frames {
+			t.Error("Did not decode expected number of frames: ", res.Decoded.Frames)
+		}
+		if 0 == res.Encoded[0].Frames {
+			// not sure what should be a reasonable number here
+			t.Error("Did not encode any frames: ", res.Encoded[0].Frames)
+		}
+	}
+
+	// Also test:
```
j0sh, line 1558: Would be good to have these tests here as well.

ffmpeg/lpms_ffmpeg.c
```diff
@@ -955,6 +965,24 @@ static int open_input(input_params *params, struct input_ctx *ctx)
...
+    return ret;
+}
+
+static int lpms_receive_frame(struct input_ctx *ictx, AVCodecContext *dec, AVFrame *frame)
+{
+    int ret = avcodec_receive_frame(dec, frame);
+    if (dec != ictx->vc) return ret;
+    if (!ret && frame && frame->pts != -1) {
+      ictx->pkt_diff--; // decrease buffer count for non-sentinel video frames
+      if (ictx->flushing) ictx->sentinel_count = 0;
```
j0sh, line 981: Curious, when will this condition be true?

ffmpeg/lpms_ffmpeg.c
jailuthra: I think so because that was the only case for which using only `pkt_diff` failed for me. We can still keep `sentinel_count` around just to be safe.

ffmpeg/lpms_ffmpeg.c
jailuthra: Setting it to `dur` instead of keeping it as-is lead to some errors in some testcases, I'll check again to be sure.

ffmpeg/lpms_ffmpeg.c
jailuthra: which one? 

`if (ictx->flushing) ictx->sentinel_count = 0;`
this is used to reset the sentinel_count counter if we are in flushing mode and received a valid frame back.

the idea is to keep increasing sentinel count if we're flushing and not receiving any valid frames back. once it hits SENTINEL_MAX we call it quits. if at any point we get a valid frame back we reset the counter.

ffmpeg/ffmpeg_test.go
jailuthra: I agree. This PR is already covering a lot of things though, will it be OK if I cut a separate issue for this?

ffmpeg/lpms_ffmpeg.c
jailuthra: Ah it is a no-op, as we are rescaling duration only when fps filter is used, and line 1229 is never executed in that case.

ffmpeg/ffmpeg_test.go
jailuthra: they do belong in `(decoded frames/2) ± 1` will that be a better check along with checking for zero?

ffmpeg/ffmpeg_test.go
j0sh: Would rather incorporate those tests now, because if the tests surface issues then we can address them prior to merging this PR. Especially because the changes here are relatively fraught, so I'd prefer to be certain that the behavior is as expected. Also, from experience, punting on tests usually means they'll never get done until we're neck deep in a bug that the test would have caught.

ffmpeg/lpms_ffmpeg.c
j0sh: > We can still keep sentinel_count around just to be safe.

The less state we have to explicitly track, the better. But it's fine for now if it maintains the current behavior.

Speaking of which, how was `TestTranscoder_MismatchedEncodeDecode ` failing? The test is set up such that if #155 is ever fixed, then the test will fail; see https://github.com/livepeer/lpms/blob/d722afc1dfa0e7b7102db8b67588a00be6dfd938/ffmpeg/ffmpeg_test.go#L1151-L1154 .

jailuthra: The testcase itself was not failing, sorry if that wasn't clear. What happened was that the decoder never returned/exited from flushing mode and went into an infinite loop of sending sentinel packets (because `pkt_diff == 1` and never reached 0)

So now we will not care about `pkt_diff` and force quit if we've sent 5 sentinel packets without getting any valid frames back.

ffmpeg/lpms_ffmpeg.c
j0sh: > this is used to reset the sentinel_count counter if we are in flushing mode and received a valid frame back.
>
> the idea is to keep increasing sentinel count if we're flushing and not receiving any valid frames back. once it hits SENTINEL_MAX we call it quits. 
>

Gotcha, so this allows us to accommodate cases where we might have to flush more than SENTINEL_MAX (eg, to continue draining the decoder), but that's okay as long as we're still receiving valid frames from the decoder?

Just curious, is there ever a case where we hit SENTINEL_MAX? Would that be a problem that the user might want to be notified of? If so, would it be helpful to return an error specifically for this case (from `send_first_pkt`?), rather than trying to "walk backwards" and infer the issue from some other failure?

ffmpeg/lpms_ffmpeg.c
```diff
@@ -1021,18 +1049,19 @@ int process_in(struct input_ctx *ictx, AVFrame *frame, AVPacket *pkt)
...
-
-    ret = avcodec_receive_frame(ictx->vc, frame);
+    ret = lpms_receive_frame(ictx, ictx->vc, frame);
     pkt->stream_index = ictx->vi;
-    if (!ret) {
-      if (is_flush_frame(frame)) ictx->flushed = 1;
-      return ret;
-    }
+    // Keep flushing if we haven't received all frames back but stop after SENTINEL_MAX tries.
+    if (ictx->pkt_diff != 0 && ictx->sentinel_count <= SENTINEL_MAX) return 0;
```
j0sh, line 1062: What if `ret` is nonzero? And we're still returning 0 here?

ffmpeg/lpms_ffmpeg.c
```diff
@@ -955,6 +965,24 @@ static int open_input(input_params *params, struct input_ctx *ctx)
...
+    int ret = avcodec_send_packet(dec, pkt);
+    if (ret == 0 && dec == ictx->vc) ictx->pkt_diff++; // increase buffer count for video packets
+    return ret;
+}
+
+static int lpms_receive_frame(struct input_ctx *ictx, AVCodecContext *dec, AVFrame *frame)
+{
+    int ret = avcodec_receive_frame(dec, frame);
+    if (dec != ictx->vc) return ret;
+    if (!ret && frame && frame->pts != -1) {
```
j0sh, line 979: `frame->pts != 1` can be replaced with `!is_flush_frame(frame)`

ffmpeg/lpms_ffmpeg.c
j0sh: Ah gotcha, that also answers the question I had here https://github.com/livepeer/lpms/pull/189#discussion_r423216120 . Thanks!

ffmpeg/ffmpeg_test.go
j0sh: > I guess only checking for non-zero frames should be good enough for now? 

Yep, good enough for now. Correct me if I'm wrong, but ffmpeg also fails to encode anything for the few-frame cases, so we can't easily sanity check these cases with ffmpeg cli?

ffmpeg/lpms_ffmpeg.c
jailuthra: >  so this allows us to accommodate cases where we might have to flush more than SENTINEL_MAX

Yep!

> is there ever a case where we hit SENTINEL_MAX

It will happen if there are no frames left to decode but our `pkt_diff` is not 0 for some reason (like #155). If we choose to return an error here instead of simply ignoring and moving on, we will break how #155 worked before this PR. Or did you have something else in mind? Logging it as a warning on stdout sounds ok, but I don't see us logging anything rn in the Cgo code.

IMO this is not an error but just a fail-proof to stop our decoder going into an infinite loop if there is any mismatch in `pkt_diff`. For example if we used ffmpeg cli or NULL flushing it would happily output mismatched frames without informing the user about it.

ffmpeg/ffmpeg_test.go
jailuthra: Yep ffmpeg fails for our 1-frame case unless we're using `eof_action=pass`. I'll test if we can possibly compare with that.

ffmpeg/lpms_ffmpeg.c
jailuthra: Ah, good catch.

So I was returning 0 here because `avcodec_receive_frame` has return value as EAGAIN for our sentinel packets, which we want to ignore for SENTINEL_MAX tries (else we kill the decoder with the error)

I've added a condition here now to only return 0 for EAGAIN and not for some other error.

ffmpeg/lpms_ffmpeg.c
j0sh: > IMO this is not an error but just a fail-proof to stop our decoder going into an infinite loop if there is any mismatch in pkt_diff

Yes, this makes sense now - we might still want to return the frames that we've already received, rather than erroring out and discarding the work that's already been done. (I originally thought this case might only be triggered with very low frame counts, maybe if *nothing* had been output, but that doesn't seem to be the case.) Thanks for the explanation!

ffmpeg/ffmpeg_test.go
jailuthra: I have updated the commit with tests for all the cases.

>  //  stream copy (both in conjunction with transcoding and standalone)

I did not exactly understand what the difference here was. I've implemented the stream copy and drop tests within the ShortSegments one after closing/re-opening the transcoder.

ffmpeg/api_test.go
```diff
@@ -478,3 +478,168 @@ func TestTranscoder_API_AlternatingTimestamps(t *testing.T) {
...
+		if err != nil {
+			t.Error("Could not transcode: ", err)
+		}
+		// verify that output frame count is same as input frame count
+		if res.Decoded.Frames != fc || res.Encoded[0].Frames != fc {
+			t.Error("Did not get expected frame count; got ", res.Encoded[0].Frames)
+		}
+	}
+
+	// test low fps (25) to low fps (20)
```
j0sh, line 605: I was thinking more like 1-2 fps as being "low fps", rather than 20/25 ? Do you think there'll be a difference?

ffmpeg/nvidia_test.go
```diff
@@ -764,4 +764,11 @@ func TestNvidia_API_AlternatingTimestamps(t *testing.T) {
 	tc.StopTranscoder()
 }
 
+func TestNvidia_ShortSegments(t *testing.T) {
+	shortSegments(t, Nvidia, 1)
+	shortSegments(t, Nvidia, 2)
+	shortSegments(t, Nvidia, 3)
+	shortSegments(t, Nvidia, 5)
```
j0sh, line 771: The pipeline is ~5 or 6 frames deep, isn't it? Maybe also add 6 and 10 frames just to sanity check and ensure the results hold? (If the results don't hold, then maybe just make a note of it in the code and we can move on)

ffmpeg/ffmpeg_test.go
j0sh: Thank you - the new tests look great!

> I did not exactly understand what the difference here was.

The code paths are a bit different if the user is doing stream copy and transcoding in different renditions.

For the streamcopy-only (or streamcopy+drop), the decoder and encoder won't be initialized at all. The packet is demuxed then sent straight to the muxer.

If doing streamcopy for one output, and transcoding another output, then the normal flow is basically in place, but the copied stream will also skip the encoder.

I thought it might be good to sanity check there weren't any weird interactions when mixing those types of outputs. We do have `TestTranscoder_StreamCopy` but that's more of a regular, single-segment test case with a normal length and fps video.

Something like this wild config could exercise a number different things at once. What do you think?

```golang
profile_low_fps = P144p30fps16x9
p.Framerate = fc // use the input frame count as the output fps, why not
profile_passthrough_fps = P144p30fps16x9
p.Framerate = 0
[]TranscodeOptions{{
  Oname: dir+"/low_fps.ts"
  Profile: profile_low_fps,
  AudioEncoder: ComponentOptions{Name:"copy"},
},{
  Oname: dir+"/copy_all.ts",
  VideoEncoder: ComponentOptions{Name:"copy"},
  AudioEncoder: ComponentOptions{Name:"drop"},
},{
  Oname: dir+"/passthrough.ts",
  Profile: profile_passthrough_fps,
}}
```

(We should also keep the standalone tests that have already been written here, because those are very good to have.)

ffmpeg/ffmpeg_test.go
j0sh: Any luck in comparing some of the cases  with ffmpeg CLI and `eof_action=pass` ?

jailuthra: Nope they varied by +-1 for almost half the non-single-frame cases

ffmpeg/ffmpeg_test.go
jailuthra: this is very wacky, i love it, added it.

ffmpeg/api_test.go
jailuthra: for <8fps ffmpeg was unable to create the input file. so i've done 8 fps -> 2 fps.

ffmpeg/nvidia_test.go
jailuthra: Done, results still hold!

ffmpeg/api_test.go
j0sh: > for <8fps ffmpeg was unable to create the input file.

Whoa. Do you think that's a bug with ffmpeg? (Not to dig into the issue, just wondering if you had any guesses as to why that might be.)

> so i've done 8 fps -> 2 fps.

👍

ffmpeg/api_test.go
```diff
@@ -478,3 +478,209 @@ func TestTranscoder_API_AlternatingTimestamps(t *testing.T) {
...
+	}
+
+	// test low fps (8) to low fps (2)
+	tc.StopTranscoder()
+	tc = NewTranscoder()
+
+	cmd = `
+    # convert segment to 8fps
+	frame_count=%d
+	ffmpeg -loglevel warning -i test.ts -vf fps=8/1 -c:v libx264 -c:a copy test8.ts
```
j0sh, line 612: I think this will encode the entire `test.ts` file when we end up only using the first $frame_count frames? Do we still get equivalent results if the two ffmpeg calls are combined? Something like:

```
ffmpeg -loglevel warning -i test.ts -vf fps=8/1 -c:v libx264 -c:a copy -frames:v $frame_count short8.ts`
```

ffmpeg/api_test.go
```diff
@@ -478,3 +478,209 @@ func TestTranscoder_API_AlternatingTimestamps(t *testing.T) {
...
+				Oname:   fmt.Sprintf("%s/passthru%d.ts", dir, i),
+				Profile: profile_passthrough_fps,
+				Accel:   accel,
+			},
+		}
+		res, err := tc.Transcode(in, out)
+		if err != nil {
+			t.Error(err)
+		}
+		if res.Encoded[0].Frames == 0 || res.Encoded[1].Frames != 0 || res.Encoded[2].Frames != fc {
```
j0sh, line 671: Checking for `res.Encoded[0].Frames == 0 ` makes me sweat a bit, because we don't really know what might be a reasonably correct value. But I understand it's pretty hard to get predictable results across different runs, probably OK for now.

ffmpeg/api_test.go
jailuthra: Yep same results, and much faster, thanks

ffmpeg/api_test.go
jailuthra: Ok I don't know what I was doing wrong before but with the new single-command conversion it works. Using 3fps -> 1fps now.

PR #194, opened 2020-05-28
Fix processing of consecutive MP4 segments

j0sh:
Fixes a number of issues around handling consecutive MP4 segments.

Closes https://github.com/livepeer/go-livepeer/issues/1514

ffmpeg: Fix decoding consecutive MP4s. … 9e0d589
A new demuxer has to be initialized for each non-mpegts input.
Preserving the old behavior saves 1% of performance.

ffmpeg: Short-circuit flushing + demuxer reset. … 8746b09
No longer have to force a read until EOF if exiting early.
Does not change anything, but it's a better way of doing things.

ffmpeg: Apply muxer options to consecutive segments with GPU transcoding … 16f55a1
For VideoProfile-directed mp4 outputs, this resulted in trailing moov.

### DIFF
--- a/ffmpeg/api_test.go
+++ b/ffmpeg/api_test.go
@@ -761,3 +761,51 @@ func fractionalFPS(t *testing.T, accel Acceleration) {
 func TestTranscoder_FractionalFPS(t *testing.T) {
 	fractionalFPS(t, Software)
 }
+
+func consecutiveMP4s(t *testing.T, accel Acceleration) {
+	run, dir := setupTest(t)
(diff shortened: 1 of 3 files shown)


j0sh:
Some further testing found a memory leak, so fixed in the following:

ffmpeg: Simplify initialization of input context. … 56bb57e
No functional changes.

Placed this change first in the branch, rather than the tip, so there isn't a leak in between commits. That is the reason for the force-push. No other changes. (Github's poor UI only shows commits in date created order, rather than how the branch is actually structured.)

PR #196, opened 2020-06-13
ffmpeg: Add support for H264 profiles

jailuthra:
### Why?
Refer to the parent issue in go-livepeer: https://github.com/livepeer/go-livepeer/issues/1531

### What?
* Add a new enum for Encoder profiles
* Add a field in VideoProfile struct to hold the encoder (h264) profile
* Pass the relevant AVOptions to FFmpeg to enable the profile
  * Support *Constrained High* by disabling B-frames using `-bf 0`

### Testing?
* Manually tested all profiles work as expected
* Manually tested by default no extra AVOption is passed so that encoder uses default profile (nvidia default is `main`, x264 default is `high`)
* Added unit tests for all the profiles, including B-frames verification for Baseline & Constrained High

### DIFF
--- a/ffmpeg/api_test.go
+++ b/ffmpeg/api_test.go
@@ -877,3 +877,134 @@ func consecutiveMuxerOpts(t *testing.T, accel Acceleration) {
 	run(cmd)
 
 }
+
+func TestTranscoder_EncodingProfiles(t *testing.T) {
+	encodingProfiles(t, Software)
+}
+
+func encodingProfiles(t *testing.T, accel Acceleration) {
+	run, dir := setupTest(t)
+	defer os.RemoveAll(dir)
+	cmd := `
+        cp "$1/../transcoder/test.ts" test.ts
+    `
+	run(cmd)
+
+	// First, sanity check defaults
+	for _, p := range VideoProfileLookup {
+		if p.Profile != ProfileNone {
+			t.Error("Default VideoProfile profile not set to ProfileNone")
+		}
+	}
+
+	// Encode to all H264 profiles
+	profilesMap := make(map[Profile]string)
+	for v := range ProfileParameters {
+		switch v {
+		case ProfileNone:
+		case ProfileH264Baseline:
+			profilesMap[v] = "baseline"
+		case ProfileH264Main:
+			profilesMap[v] = "main"
+		case ProfileH264High:
+			profilesMap[v] = "high"
+		case ProfileH264ConstrainedHigh:
+			profilesMap[v] = "constrained_high"
+		default:
+			t.Error("Unhandled profile ", v)
+		}
+	}
+
+	for codecProfile, profileString := range profilesMap {
+		tc := NewTranscoder()
+		profile := P144p30fps16x9
+		profile.Profile = codecProfile
+		in := &TranscodeOptionsIn{Fname: fmt.Sprintf("%s/test.ts", dir), Accel: accel}
+		out := []TranscodeOptions{{
+			Oname:   fmt.Sprintf("%s/out_%s.mp4", dir, profileString),
+			Accel:   accel,
+			Profile: profile,
+		}}
+		_, err := tc.Transcode(in, out)
+		if err != nil {
+			t.Error("Unexpected error ", err)
+		}
+		tc.StopTranscoder()
+	}
+
+	// Verify outputs have correct profile names
+	cmd = `
+		ffprobe -loglevel warning -show_streams out_baseline.mp4 | grep "profile=Constrained Baseline"
+		ffprobe -loglevel warning -show_streams out_main.mp4 | grep "profile=Main"
+		ffprobe -loglevel warning -show_streams out_high.mp4 | grep "profile=High"
+		ffprobe -loglevel warning -show_streams out_constrained_high.mp4 | grep "profile=High"
+	`
+	run(cmd)
+
+	// Verify the two constrained profiles (Baseline & Constrained High) do not have any b frames
+	cmd = `
+		ffprobe -loglevel warning -show_streams out_baseline.mp4 | grep has_b_frames=0
+		ffprobe -loglevel warning -show_streams out_constrained_high.mp4 | grep has_b_frames=0
+	`
+	run(cmd)
+
+	// Verify the other two profiles have B frames
+	cmd = `
+		ffprobe -loglevel warning -show_streams out_main.mp4 | grep has_b_frames=2
+		ffprobe -loglevel warning -show_streams out_high.mp4 | grep has_b_frames=2
+	`
+	// TODO: Only CPU encoder seems to use B frames by default, check if we can enable them for NVIDIA
+	if accel == Software {
+		run(cmd)
+	}
+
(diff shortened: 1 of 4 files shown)


ffmpeg/api_test.go
```diff
@@ -877,3 +877,79 @@ func consecutiveMuxerOpts(t *testing.T, accel Acceleration) {
...
+	// Verify outputs have correct profile names
+	cmd = `
+		ffprobe -loglevel warning -show_streams out_baseline.mp4 | grep "profile=Constrained Baseline"
+		ffprobe -loglevel warning -show_streams out_main.mp4 | grep "profile=Main"
+		ffprobe -loglevel warning -show_streams out_high.mp4 | grep "profile=High"
+		ffprobe -loglevel warning -show_streams out_constrained_high.mp4 | grep "profile=High"
+	`
+	run(cmd)
+
+	// Verify the two constrained profiles (Baseline & Constrained High) do not have any b frames
```
j0sh, line 933: Can we also check that the Main / High profiles do have B-frames in the output? (This might be tricky; IIRC one of the encoders - either x264 or cuda - did not output any B-frames while the other did, under the default configuration.)

ffmpeg/api_test.go
```diff
@@ -877,3 +877,79 @@ func consecutiveMuxerOpts(t *testing.T, accel Acceleration) {
 	run(cmd)
 
 }
+
+func TestTranscoder_EncodingProfiles(t *testing.T) {
+	encodingProfiles(t, Software)
+}
+
+func encodingProfiles(t *testing.T, accel Acceleration) {
```
j0sh, line 885: Curious, should we add an interlaced test case for constrained high? Even though we know the behavior is incorrect, it might help to codify that until we get a more permanent solution in. Something along the lines of this https://github.com/livepeer/lpms/blob/ec8e1e4e242024eaa0a7c83cfe67fe20a4d34d11/ffmpeg/ffmpeg_test.go#L1104-L1110

ffmpeg/videoprofile.go
```diff
@@ -16,6 +16,16 @@ const (
 	FormatMP4
 )
 
+type Profile int
+
+const (
+	ProfileNone Profile = iota
+	H264Baseline
```
j0sh, line 23: Should we use a naming scheme along the lines of `ProfileH264Baseline` ? Matches the `ProfileNone`, and follows the `Format*` convention around enum-like values.

ffmpeg/videoprofile.go
```diff
@@ -16,6 +16,16 @@ const (
...
+	H264Main
+	H264High
+	H264ConstrainedHigh
+)
```
j0sh, line 27: Should we add some way to actually enumerate the profiles, similar to the `FormatExtensions` table? Maybe something like a `map[Profile]bool` . (There might be something else we can use as a value rather than a bool, but I can't think of anything.) 

That helps check whether a certain int (enum) supplied is actually a valid profile. go-livepeer uses these in a few places to fail out on missing support for new formats, eg

https://github.com/livepeer/go-livepeer/blob/ed7f24c763ad8db4592db5b4d1a2f2ce83aa6ca0/common/util_test.go#L181-L183

or

https://github.com/livepeer/go-livepeer/pull/1539/files#diff-3d8fe498e1aa211419d9b8f4b696b528R132-R136

For example, this could help check that the `Transcode` function within LPMS handles all profiles, or the webhook JSON conversion function in go-livepeer.

ffmpeg/api_test.go
jailuthra: Yep I had a testcase for that and removed it precisely because NVENC did not encode any bframes even for main/high profile.

I'll add a CPU only test case for it, as I do not want to change the default behaviour of NVENC for main/high profile (even if it is possible to manually force B-frame usage through some parameter)

ffmpeg/api_test.go
jailuthra: Actually I think we can safely say the behaviour as correct. Testing further with `ffprobe -show_frames` showed me that the `interlaced_frame` flag is unset by default on all profiles, unless we [explicitly enable](https://stackoverflow.com/questions/22471099/how-to-convert-a-1080p-to-1080i-using-ffmpeg) MBAFF during encoding via `-flags +ilme+ildct -x264opts tff=1`. (The video still looks stripey if original content was interlaced regardless of if MBAFF is used or not. The only way to prevent that is still an expensive deinterlacing filter before encoding)

I will add a testcase to assert that the `interlaced_frame` parameter is unset for baseline & constrained high.

ffmpeg/videoprofile.go
jailuthra: Added a `map[Profile]string` that can also be used to find the profile parameters to pass to FFmpeg along with enumeration.

jailuthra:
Force Push: Added H264 Profile string lookup map used by the auth webhook in go-livepeer

ffmpeg/api_test.go
j0sh: Sounds good - yeah it would help compressibility to incorporate B-frames at some point, but it's not needed for now. Nvidia has some other issues too that feed into this a bit, like it'll output a constant bitrate whether it actually *needs* that many bits. Unlike x264, where the bit rate is more of a maximum, unless you explicitly enable something like cbr or vbv.

ffmpeg/videoprofile.go
```diff
@@ -79,6 +90,23 @@ var ExtensionFormats = map[string]Format{
...
 
+var ProfileParameters = map[Profile]string{
+	ProfileNone:                "",
+	ProfileH264Baseline:        "baseline",
+	ProfileH264Main:            "main",
+	ProfileH264High:            "high",
+	ProfileH264ConstrainedHigh: "high",
+}
+
+var EncoderProfileLookup = map[string]Profile{
```
j0sh, line 101: This mapping seems specific to the webhook; it feels like it belongs better within go-livepeer.

ffmpeg/api_test.go
```diff
@@ -877,3 +877,123 @@ func consecutiveMuxerOpts(t *testing.T, accel Acceleration) {
...
+
+	// First, sanity check defaults
+	for _, p := range VideoProfileLookup {
+		if p.Profile != ProfileNone {
+			t.Error("Default VideoProfile profile not set to ProfileNone")
+		}
+	}
+
+	// Encode to all H264 profiles
+	var profilesMap = map[Profile]string{
```
j0sh, line 901: It might be good to write this test in a way that it fails out if we ever add more profiles to the `ProfileParameters ` mapping. Something like this:

```golang
var profilesMap map[Profile]string
for v, _ := ProfileParameters {
  switch v {
    case ProfileNone:
      // skip
    case ProfileH264Baseline:
      profilesMap[v] = "baseline"
    ... check rest of profiles ...
    default: 
      t.Error("Unhandled profile ", v)
  }
}
```

ffmpeg/api_test.go
j0sh: I'm just waiting for the day until someone sends in interlaced content and complains about stripey output. Until then, we can say it's working as intended...

PR #198, opened 2020-06-25
ffmpeg: Knob to tune output GOP intervals.

j0sh:
Adds user control over the output GOP interval.

### Changes
* Add a "GOP" field to VideoProfile in units of `time.Duration`.  8f0ad3555d9f1f385e4d17180e6cd6c0a780383d
* If fixed FPS output is used, use built-in `g` AVOption to have the encoder output keyframes based on a fixed frame interval. 8f0ad3555d9f1f385e4d17180e6cd6c0a780383d
* If passthrough FPS is used, keep track of when the next keyframe is due based on timestamps and explicitly assign frames to keyframes within LPMS. 78548c9079617ff8c6cddb4a90cb2fafe5ddf9da
* Adds an intra mode to demonstrate support for extended GOP modes. f1ca834f064f6b052fc9a48bde72aedf3024478f  
* simplify test runner 4c6d62c9f22011e48a538c46a885397b1fde475d
* Error checking for invalid GOP values. This starts at -1 and is bumped down to -2 for intra mode. This should be bumped down further if additional modes are ever added, such as source matching.

### DIFF
--- a/ffmpeg/api_test.go
+++ b/ffmpeg/api_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"testing"
+	"time"
 )
 
 func TestAPI_SkippedSegment(t *testing.T) {
@@ -1007,4 +1008,74 @@ func encodingProfiles(t *testing.T, accel Acceleration) {
 		t.Errorf("Unexpected error; wanted %v but got %v", ErrTranscoderPrf, err)
 	}
 	tc.StopTranscoder()
+}
+
+func TestAPI_SetGOPs(t *testing.T) {
+	setGops(t, Software)
+}
+
+func setGops(t *testing.T, accel Acceleration) {
+	run, dir := setupTest(t)
+	defer os.RemoveAll(dir)
+	cmd := `
+
+        cp "$1/../transcoder/test.ts" .
+
+        # ffmpeg sanity checks
+
+        ffmpeg -loglevel warning -i test.ts -c copy -f hls out.m3u8
+        function prepare {
+            f=$1
+
+            ffprobe -loglevel warning $f -select_streams v -show_packets | grep flags=K | wc -l | grep 1
+            ffprobe -loglevel warning $f -select_streams v -show_packets | grep flags= | wc -l | grep 120
+            ffmpeg -loglevel warning -i $f -force_key_frames 'expr:if(isnan(prev_forced_t),1,gte(t,prev_forced_t+0.3))' -an -c:v libx264 -s 100x100 forced_$f
+        }
+
+        prepare out0.ts
+        prepare out1.ts
+        prepare out2.ts
+        prepare out3.ts
+
+    `
+	run(cmd)
+
+	p := P144p30fps16x9
+	p.GOP = 300 * time.Millisecond
+	tc := NewTranscoder()
+	defer tc.StopTranscoder()
+	for i := 0; i < 4; i++ {
+		fname := fmt.Sprintf(dir+"/out%d.ts", i)
+		oname := fmt.Sprintf(dir+"/lpms%d.ts", i)
+		_, err := tc.Transcode(&TranscodeOptionsIn{Fname: fname, Accel: accel}, []TranscodeOptions{{Oname: oname, Accel: accel, Profile: p}})
+		if err != nil {
+			t.Error(err)
+		}
+	}
+
+	cmd = `
+        function check {
+            f=$1
+            ffprobe -loglevel warning $f -select_streams v -show_packets | grep flags=K | wc -l | grep 7
+        }
+
+        # ensures we match ffmpeg with the same check
+        check forced_out0.ts
+        check forced_out1.ts
+        check forced_out2.ts
+        check forced_out3.ts
+
+        check lpms0.ts
+        check lpms1.ts
+        check lpms2.ts
+        check lpms3.ts
+    `
+	run(cmd)
+
+	// check invalid gop lengths
+	p.GOP = GOPInvalid
+	_, err := tc.Transcode(&TranscodeOptionsIn{Fname: "asdf"}, []TranscodeOptions{{Oname: "qwerty", Profile: p}})
+	if err == nil || err != ErrTranscoderGOP {
+		t.Error("Did not expected error ", err)
+	}
 }
--- a/ffmpeg/api_test.go
+++ b/ffmpeg/api_test.go
@@ -1037,6 +1037,10 @@ func setGops(t *testing.T, accel Acceleration) {
         prepare out2.ts
         prepare out3.ts
 
+        # extremely low frame rate
+        ffmpeg -loglevel warning -i test.ts -c:v libx264 -r 1 lowfps.ts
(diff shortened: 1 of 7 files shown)


f1l1b0x:
dont you have to give nvenc an exact number of frames? if you use a time to set it like 2 seconds how do you know what to set in nvenc?

j0sh:
Actually have not tested with nvenc yet (thanks for the reminder). But under the hood, we do set an exact number of frames: if the framerate is constant, then given a `gop_time` in seconds, then the number is exactly `fps*gop_time` .

If the framerate is passthrough, then we set a frame flag forcing the encoder to output a keyframe, based on the pts difference from the previous keyframe.

j0sh:
I am starting to lean towards converting the GOP string to use an integer in milliseconds, with the zero value ignored for now. The webhook can remain a string and we'll do the secs->ms conversion in go-livepeer. Intra-only / source-matching can be signaled in the future via special negative values (-1, -2, etc).

Any thoughts about this?

In particular, should we attempt to accommodate frame lists or a fixed frame interval (not time based) under the same signaling mechanism? ffmpeg also has a rather neat expression-evaluation mechanism [1]. This is perhaps future-proofing excessively; my feeling is those features have limited usefulness in most contexts, and specifying the GOP in time units is what most users actually want.

[1] "force key frames" in the [ffmpeg documentation](https://ffmpeg.org/ffmpeg.html#Advanced-Video-options)

iameli:
+1 to milliseconds instead of seconds if only to match the `Content-Duration` header.

With your changes, `0` for this setting matches the existing behavior, yeah? That seems good.

So we've been handling a lot of segments that are arriving to us downstream of a shaky RMTP input with lots of frame drops. In that situation, one thing that seems to be happening a lot is that we _only_ get keyframes from the incoming stream. (I don't know if this is OBS's behavior or Mist's or what, but I've seen it a lot.) So... what happens in this scenario?

1. Source has a 3s keyframe interval
2. Connection is shaky, all non-keyframes are dropped
3. Livepeer is sent a segment containing two keyframes, with a PTS 3000ms apart. My `gop` parameter is set to `2000`.

So the source segment is like:

`[Keyframe A] [3000 ms PTS gap] [Keyframe B]`

Is the return segment the same? Or do we double-up on the frames to enforce the GOP interval, resulting in...

`[Keyframe A] [2000 ms PTS gap] [Copy of Keyframe A] [1000 ms PTS gap] [Keyframe B]`?

Or even like...

`[Keyframe A] [2000 ms PTS gap] [Copy of Keyframe A] [1000 ms PTS gap] [Keyframe B but it's an I-Frame now]`?

To be clear, I don't think there's any particularly coherent way for us to respond in that situation, I just know it's gonna happen so I want to wrap my head around it.

j0sh:
>  if only to match the Content-Duration header.

So the go-livepeer webhook should be in milliseconds as well?

I was thinking of stringifying the webhook value, while maintaining things in integer milliseconds internally. Only because I can easily see scenarios where the GOP could be overloaded - eg, expressed in seconds (eg, `gop:"2.5"`), or source-matched (eg, `gop: "source"`), or intra-only (eg, `gop:"intra"`), or maybe even a list of frames (eg, `gop:"1,3,5,7,9"`).

If we ever need to support other types of gop specifiers, do you think we should have separate webhook fields for these? Eg, `GopSourceMatching: true` ?

Also, somewhat related aside, but reflecting back on the FPS work, it feels like that type of overloaded string scheme might have been a better fit for the webhook, had we originally specified the fps value as a string, rather than a uint. Could have offered a fractional string (`fps:"30000/1001"`), or a decimal (`fps:"29.97"`) or even a named value (`fps:"ntsc"`), in addition to a round integer (`fps:"30"`)

> With your changes, 0 for this setting matches the existing behavior, yeah?

Whew:

* To match the existing behavior within this PR, use the empty string "" . This is the default when no GOP is specified. 
* If a "0" is passed in, that results in intra-only frames because that's what ffmpeg does when you set the gop size to 0. Only applies with fixed fps output framerates.
* If passthrough fps is used, "0" is ignored, and it matches the default behavior. Just because that's how it works right now (have a local unit test checking for this condition).

Writing all this out is just me confirming to myself that this needs to be made more consistent. In particular, "0" == "intra-only" feels a bit error-prone, even though it is basically logically correct.

> [Keyframe A] [3000 ms PTS gap] [Keyframe B]
>
> Is the return segment the same? 
>

That depends on the fps mode.

If a fixed fps output is used (say, 30 fps) then the result would have around 90 frames, all duplicates of the first, except for the last one. Keyframes would be at the first frame and at the 2000ms point. The last frame, Keyframe B, would likely be a regular, non-intra frame, but it depends on the encoder behavior.

If passthrough fps is used, then only those two frames would be output, and both would be encoded as keyframes.

> I don't think there's any particularly coherent way for us to respond in that situation,

Indeed. Garbage in, garbage out.

jailuthra:
> Allows specifying a gop of "0" for intra-only without conflating with golang's zero value. (This matches ffmpeg's behavior.)

I can see that the internal `gop_size` field by FFmpeg is used similar to this (i.e. set to 0 means intra-only mode)
But from `$ ffmpeg -h full`: `-intra              deprecated use -g 1`

I think it would be a good idea to test which of the two works more reliably for x264 and nvenc `-g 0` or `-g 1`, and maybe add a unit test that performs a sanity check of intra-only frame-by-frame?

Other than that LGTM!

> Also, somewhat related aside, but reflecting back on the FPS work, it feels like that type of overloaded string scheme might have been a better fit for the webhook

Yeah I thought we didn't go that way to maintain back-compat with the uint fps field for webhook before the fractional changes.

> In particular, "0" == "intra-only" feels a bit error-prone, even though it is basically logically correct.

I think this is fine. The only point of confusion is that 0 is usually default for int fields, but given we're using a string field, I think just making it clear that "" is default is good enough.

I presume the usecase for intra-only frames would be extremely rare anyway?

> This is perhaps future-proofing excessively; my feeling is those features have limited usefulness in most contexts

I agree. We can explore the other `force_key_frames` option later if there is a demand for something like it. FFmpeg options are horribly documented and don't work with every codec [[1]](https://superuser.com/questions/908280/what-is-the-correct-way-to-fix-keyframes-in-ffmpeg-for-dash) so the future proof thing here is to actually avoid messing unless needed :stuck_out_tongue:

iameli:
Arg, I definitely typed up a response to this on Friday but didn't hit comment or something. @j0sh string seems totally fine in the webhook, agreed that I kinda wish we had gone with more strings there, we could do stuff like `{"height": "1080", "width": "contain"}`. Or whatever. This all seems good.

j0sh:
> Yeah I thought we didn't go that way to maintain back-compat with the uint fps field

Indeed, that's why. 

> I presume the usecase for intra-only frames would be extremely rare anyway?

Most likely, yes. Matching the source GOP is probably going to be requested more often. Neither are a requirement right now, but I think I'll put in intra-only just to make the behavior consistent between fixed and passthrough fps, unless we explicitly reject zero-GOP (or 1-GOP) intervals.

j0sh:
Ready for final review. PR Description has been updated with details of what has changed.

Issue #199, opened 2020-06-29
Robust timestamp handling


jailuthra:
## Abstract

Currently the LPMS transcoder faces a lot of issues related to segment ordering and frame timestamps. This is a proposal to make the transcoding process more robust to failure when input segments are out of order or are missing a few frames.

## Motivation

Edge cases of out-of-order segments and missing frames happen often in real use of LPMS/Livepeer nodes. For example while retrying after a timeout/failure the Broadcaster would send a segment again which might have overlaping timestamps of a previously transcoded segment and thus fail. Or sometimes the client's mediaserver will drop some frames and thus skip over a chunk of timestamp, causing time drift in playback. (See #180 and #197 for more details)

## Proposed Solution

```
                                               FILTERGRAPH
                         +------------------------------------------------------+
                         |                                                      |
                         |      +------------+              +------------+      |
                         |      |            |              |            |      |
                         |      |            |              |            |      |
  +-----------+          |      |  SCALE     |              |  FPS       |      |         +-----------+
  |           |          |      |   filter   |              |   filter   |      |         |           |
  | decoded F +---------------->+            +------+------>+            +--------------->+ encoded F |
  |           |          |      |            |      ^       |            |      |         |           |
  +-----------+          |      |            |      |       |            |      |         +-----------+
                         |      |            |      |       |            |      |
     pts_in              |      |            |      |       |            |      |             pts_out
                         |      +------------+      |       +------------+      |    (2)
(non-monotonic &         |                          |                           |    (guess using pts_in & fps
 unreliable jumps)       +------------------------------------------------------+       to maintain same order)
                                                    |
                                                    |
                                                    |
                                                    +
                                           (1) dummy monotonic
                                                   pts
```

1. FPS filter expects monotonic increase in input frame's PTS. We cannot rely on the input to be monotonic thus:
    i. Set a dummy PTS before the frame is sent into filtergraph, that we manually increase monotonically.
    ii. OR use *SETPTS* filter in the filtergraph before FPS filter, which would do the same thing.

2. If the input had missing frames (jumps in PTS) or if we had used dummy PTS before for the fps filter - we would need to set the encoded frame's PTS manually to ensure correct order and timescaling after change in FPS.

### 1. Force monotonic PTS before FPS filter

We have two possible approaches here. Josh's WIP patch definitely works. Need to verify the `setpts` filter approach first, even though it looks cleaner and worked well with the ffmpeg cli.

**i. Josh's WIP patch**

[...]

Issue #197, opened 2020-06-25
Gaps after dropped segments are ignored; timestamp drifts over time


iameli:
This occurs when we're processing data that contains missing chunks, like when we're handling data from MistServer RTMP ingest and OBS looks like this:

![image](https://user-images.githubusercontent.com/257909/85639958-349b2880-b63f-11ea-96a8-f879b75380e5.png)

As far as I understand it, when we get to those dropped segments, we incorrectly treat the next frame we encounter "next" in the timecode. This results in the transcoded PTS drifting behind the source PTS over a long enough time period, here's a spreadsheet of it happening: https://docs.google.com/spreadsheets/d/1UzP5EmMuFF6rw-wlOg2vIX1fT6XPa-hto0w_qWrEkvA/edit#gid=0

Minimal repro case; transcode these three segments in sequence and then look at the discrepancies between the start and end: [sample.tar.gz](https://github.com/livepeer/lpms/files/4828510/sample.tar.gz)

Full 1GB dataset: https://storage.googleapis.com/docs.livepeer.live/eli/segment-debugging/drift.tar.gz

EDIT: We think this to be caused by changing FPS

jailuthra:
x-posting some root cause analysis on this from discord:

The source segments seem to have some missing frames - there were weird jumps in source PTS around the 1-frame segments. The current lpms transcoder by default assumes there are no missing frames in the input, and predicts output PTS uniformly with every incoming frame (regardless of what incoming frame says its PTS is)

Thus output PTS ultimately drifts exactly the same amount as the duration of missing frames or pts jumps in the input. Analyzing timestamps of all the source-xxx.ts files in the shared tar, the drift is around 77seconds :O

Here's a sheet with the frame-wise PTS data: https://docs.google.com/spreadsheets/d/1v966abKvmHoVbWrjot1wY5HOfNYsHfCTgyF86llmplY/edit#gid=0

Here's the short bash command I used to generate it:
```for i in {376..1967}; do; ffprobe -show_frames source-$i.ts | grep pkt_pts_time >> pts.txt; done;```

*Note to future solver (probably me):* The fix would be to handle setting output PTS robustly even if source segments have dropped frames in between. Find the places where LPMS assumes happy-paths wrt source frame PTSs.

PR #201, opened 2020-07-24
ffmpeg: Custom timestamp handling

jailuthra:
**Why?**
Look at #199 for details on what this PR fixes.

**What has been changed?**

* Disable check for Out-Of-Order segments
* Set uniformly-increasing monotonic PTS as the input to filtergraph (when FPS filter is used)
* Set `opaque` to original PTS
* Use original PTS rescaled to the filtergraph output-timebase to calculate the correct output PTS
  * This is done only for the first frame of the segment
  * Store the difference between the output PTS filtergraph sent for the first frame, and the calculated PTS for the first frame of the segment
  * Apply this difference on every subsequent frame (This ensures uniformly increasing output for a particular segment, but maintains the original intra-segment order)

For example see the output of the AlternatingTimestamps test case [here](https://gist.github.com/jailuthra/784cc1c30d1f0eeb44b24676f0b399b4#file-lpms_alternating_segments-txt). The segments are out of order - `1,0,3,2`. The output shows different PTS values in the order - `original_pts, rescaled_to_output, filterbuffer_output, final_calculated_pts`.

The final calculated PTS is now uniformly increasing, and maintains intra-segment ordering by being close to the rescaled value of the original input PTS.

**Testing**

* Manually tested `transcoding` binary produces usual output with correct timestamps
* Manually tested out of order segments usecase
* Modified AlternatingTimestamps unit test to run completely
* Manually tested the segments in #197 with missing frames, they work without causing drift now.
* ~~**WIP** Disabled or tweaked other failing test cases for now until the approach is verified~~
* Modified other test cases as needed, to match the new approach (some 1-on-1 pts comparisons with ffmpeg fail now but that was expected)
* ~~TODO~~ Added a custom testcase to reproduce #197 using the sample segment present in LPMS

### DIFF
--- a/ffmpeg/api_test.go
+++ b/ffmpeg/api_test.go
@@ -1129,3 +1129,53 @@ func setGops(t *testing.T, accel Acceleration) {
 		t.Error("Did not expected error ", err)
 	}
 }
+
+func TestTranscoder_SkippedFrames(t *testing.T) {
+	// Reproducing #197
+	run, dir := setupTest(t)
+	defer os.RemoveAll(dir)
+	cmd := `
+    cp "$1/../transcoder/test.ts" .
+    ffmpeg -loglevel warning -i test.ts -c:a copy -vf fps=30 -c:v libx264 -muxdelay 0 test30.ts
+    ffmpeg -loglevel warning -i test30.ts -vf select='between(n\,0\,49)' -c:a copy -c:v libx264 -muxdelay 0 -copyts source-0.ts
+    ffmpeg -loglevel warning -i test30.ts -vf select='between(n\,60\,62)' -c:a copy -c:v libx264 -muxdelay 0 -copyts source-1.ts
+    ffmpeg -loglevel warning -i test30.ts -vf select='between(n\,120\,179)' -c:a copy -c:v libx264 -muxdelay 0 -copyts source-2.ts
+    for i in source-*.ts
+    do
+      ffprobe -show_streams -select_streams v $i | grep start_time >> source.txt
+    done
+  `
+	run(cmd)
+
+	tc := NewTranscoder()
+	defer tc.StopTranscoder()
+
+	prof := P144p30fps16x9
+	prof.Framerate = 45
+	// Test encoding
+	for i := 0; i < 3; i++ {
+		in := &TranscodeOptionsIn{
+			Fname: fmt.Sprintf("%s/source-%d.ts", dir, i),
+			Accel: Software,
+		}
+		out := []TranscodeOptions{{
+			Oname:   fmt.Sprintf("%s/out-%d.ts", dir, i),
+			Profile: prof,
+			Accel:   Software,
+		}}
+		_, err := tc.Transcode(in, out)
+		if err != nil {
+			t.Error(err)
+		}
+	}
+
+	// Compare start times of original segments to the transcoded segments
+	cmd = `
+    for i in out-*.ts
+    do
+      ffprobe -show_streams -select_streams v $i | grep start_time >> out.txt
+    done
+    diff -u source.txt out.txt
+  `
+	run(cmd)
+}
--- a/ffmpeg/api_test.go
+++ b/ffmpeg/api_test.go
@@ -17,6 +17,7 @@ func TestAPI_SkippedSegment(t *testing.T) {
 	// tests that also check things like SSIM, MD5 hashes, etc...
 	// See TestNvidia_API_MixedOutput / TestTranscoder_EncoderOpts / TestTranscoder_StreamCopy
 	run, dir := setupTest(t)
+	defer os.RemoveAll(dir)
 	err := RTMPToHLS("../transcoder/test.ts", dir+"/out.m3u8", dir+"/out_%d.ts", "2", 0)
 	if err != nil {
 		t.Error(err)
@@ -277,12 +278,12 @@ pkt_pts=847500
 pkt_pts=848250
 
 ==> out_30fps_0.ts.pts <==
-pkt_pts=129000
 pkt_pts=132000
 pkt_pts=135000
-pkt_pts=300000
+pkt_pts=138000
 pkt_pts=303000
 pkt_pts=306000
+pkt_pts=309000
 
 ==> out_30fps_1.ts.pts <==
 pkt_pts=309000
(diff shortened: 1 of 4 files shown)


ffmpeg/api_test.go
```diff
@@ -403,6 +403,7 @@ func TestTranscoder_API_AlternatingTimestamps(t *testing.T) {
 	// tests that also check things like SSIM, MD5 hashes, etc...
 	// See TestNvidia_API_MixedOutput / TestTranscoder_EncoderOpts / TestTranscoder_StreamCopy / TestNvidia_API_AlternatingTimestamps
 	run, dir := setupTest(t)
+	defer os.RemoveAll(dir)
```
j0sh, line 406: Oops. At some point, it might be nice to switch to go 1.14 and make use of the `test.Cleanup` helper which can be invoked within `setupTest`, so we don't have to worry about this bit of boilerplate everywhere. See https://www.gopherguides.com/articles/test-cleanup-in-go-1-14/

ffmpeg/ffmpeg_test.go
```diff
@@ -1073,6 +1073,7 @@ func TestTranscoder_RepeatedTranscodes(t *testing.T) {
 }
 
 func TestTranscoder_MismatchedEncodeDecode(t *testing.T) {
+	// ALL FIXED, Keeping it around as a sanity check for now
```
j0sh, line 1076: Woot! 🎉

j0sh:
The general approach looks good!

ffmpeg/lpms_ffmpeg.c
```diff
@@ -1173,39 +1176,64 @@ int process_out(struct input_ctx *ictx, struct output_ctx *octx, AVCodecContext
...
+    if (inf) {
+      inf->opaque = (void *) inf->pts;
+      filter->custom_pts = inf->pts;
+    } else if (!filter->flushed) {
+      // Audio Flushing
+      AVFrame *frame = ictx->last_frame_a;
+      filter->custom_pts += frame->pkt_duration;
+      inf = frame;
+      inf->opaque = NULL;
+      is_flushing = 1;
```
jailuthra, line 1227: Ah now that I read this, it can be simplified a lot. Will do it in next iteration.

j0sh:
Wondering if you've also looked at modifying the FPS filter to accommodate this type of behavior? A while ago we experimented with adding a "flush" command. That could do two things here:

* Return all pending frames
* Reset the starting timestamp, so the next frame that comes in is "first"

The latter is what could help us accommodate out-of-order segments. It'd be kind of specific to our use-case, and might have to be carried as an out-of-tree patch indefinitely. But that might be preferable if it means we can clean up all this bookkeeping around filter timestamp handling.

jailuthra:
>     * Return all pending frames
@j0sh Yep this already worked with my out-of-tree patch

>     * Reset the starting timestamp, so the next frame that comes in is "first"
> The latter is what could help us accommodate out-of-order segments. It'd be kind of specific to our use-case, and might have to be carried as an out-of-tree patch indefinitely. But that might be preferable if it means we can clean up all this bookkeeping around filter timestamp handling.

I agree! My patch didn't solve the latter, and it seemed tricky to do it with minimal change in the filter (other than closing the context completely).

But I totally think that we can probably write our own version of fps filter (the ffmpeg one is ~300 lines) that is much simpler and matches our exact segmented-workflow usecase. In fact, maintaining a full patchset or FFmpeg fork is one of the alternatives if we decide to stick to segment-based workflow. We are currently trying to fool the ffmpeg api by book-keeping which makes things very hard.

I thought for now it would be good to stick with the current approach in this PR until the segment v/s contiguous idea is clearer. What do you think?

j0sh:
> I thought for now it would be good to stick with the current approach in this PR until the segmented v/s contiguous future is clearer. What do you think?

Totally up to you. If a continuous approach is chosen, there is a further dependency on the the design / implementation plan and any perceived risks associated with that. The current approach has a higher maintenance burden but mostly works thanks to this PR. If the sense is that the segmented flow will be replaced sooner rather than later, then it makes sense to keep things as-is. However, if there's a possibility that the segmented flow will remain for some time - and personally, I think it will take a while to fully shift to a continuous approach (or it may be that both approaches need to be supported simultaneously for a while) - then it seems better to use the custom filter for now. Good luck!

ffmpeg/lpms_ffmpeg.c
```diff
@@ -115,8 +115,9 @@ struct filter_ctx {
   //      ( pts received from filter will be wrong by however many flushed
```
yondonfu, line 115: Is this comment block still accurate given the new timestamp handling approach?

ffmpeg/ffmpeg_test.go
```diff
@@ -1175,6 +1176,7 @@ func TestTranscoder_FFmpegMatching(t *testing.T) {
 	// 1 second input, N fps ( N frames )
 	// Output set to M fps. Output contains M frames
 
+	// TODO Unable to compare these since diverged FPS handling from from FFmpeg, Find a better way
```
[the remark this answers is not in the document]
jailuthra: Ah yes will move this up, thx!

ffmpeg/lpms_ffmpeg.c
jailuthra: Good catch, completely ignored the comments in this struct while making changes. Will remove thx.

jailuthra:
> Did the `setpts` filter alternative approach described in #199 end up not working?

@yondonfu Ah yes I mentioned this during a standup but didn't update the spec. https://discord.com/channels/423160867534929930/646028985951715368/728212353661010021

To use setpts we needed to hardcode the input FPS while [initializing the filtergraph](https://github.com/livepeer/lpms/blob/master/ffmpeg/ffmpeg.go#L216) in the Go module. Something like `setpts=N/(25*TB)` for uniform input PTS at 25fps. Calculating the input FPS manually there is a hard task (as the incoming segment may report wrong FPS in the metadata or have too few frames) so it is best left to be done by the FFmpeg demuxer+decoder later in the Cgo module (which we do now).

The `setpts` [filter](https://ffmpeg.org/ffmpeg-filters.html#setpts_002c-asetpts) does support an in-built option called `FRAME_RATE/FR` but that again was often coming out to be `0` as it required the input stream to have constant framerate defined in the bitstream headers.

PR #204, opened 2020-08-20
Handle zero-frame (audio-only) segments at the start of a session

### LINKED ISSUE #203: Handle audio-only segments
### Current Behavior

Sometimes Livepeer is sent audio-only segments. This could happen for a variety of reasons. In the case of MistServer, it happens when a transcode is enabled and there is more audio data in the buffer than video data. It could also happen in a variety of edge cases where bandwidth is severely constrained and end-user client software decides to only send audio to keep _something_ online. (I'm especially here thinking about livestreams from protests and whatnot — audio is better than nothing!)

Currently, we barf back a not-very-descriptive error in these cases:

```
orchestrator.go:520] Error transcoding manifestID=redacted segNo=1 segName= - Unsupported input pixel format 
segment_rpc.go:174] Could not transcode manifestID=redacted seqNo=1 err=Unsupported input pixel format 
```

This error gets bubbled back up to the broadcaster and incoming HTTP ingest segments get rejected. Instead of a contiguous audio stream with temporarily paused video, viewers see nothing until the next video segment arrives. In the MistServer case, Mist retries the transcode several times on several broadcasters and eventually gives up on the stream session entirely, causing a break in the recording if that's enabled. No good!

### Desired Behavior

Instead of erroring on such segments, let's accept them and send back audio-only tracks. The audio output should be exactly the same as the audio input. We handle one-frame segments great these days... now we just need to handle zero-frame segments.

### Concerns

Presumably the pixel count of an audio-only segment is zero. I'm slightly worried about a DoS attack against orchestrators here? If I spam a public orchestrator with high-bitrate audio, could I saturate their bandwidth and take down the orchestrator without paying the orchestrator at all? Concerning. Perhaps a cap on audio bitrate?

### Alternatives

I think the dream is that such segments could be handled on the _broadcaster_, rather than the transcoder. It's not as if the OT is actually doing anything in this case. But I imagine that would require a lot more video introspection on the B than we're currently doing, so I figure this is a quicker fix.

jailuthra:
**Why?**
For details check the audio-only bug this PR fixes #203

**How?**
We bypass the usual transcoding process for the first n-segments of a stream that do not have any frames, but do have a valid *video substream* present in the container format.
The bypassing is done by forcing the copy transcoder https://github.com/livepeer/lpms/commit/8bc28e3f702049a17c24ab2041857a47d8af51bf for such segments.

**Testing Done:**
- Manually reproduced with a failing segment from prod
- Reproduced the issue with a unit test https://github.com/livepeer/lpms/commit/c372f562194ea0c88595d7f05797f827f6ffd683
- Fixed & verified the fix for both of the above ^

### DIFF
--- a/ffmpeg/api_test.go
+++ b/ffmpeg/api_test.go
@@ -1175,3 +1175,69 @@ func TestTranscoder_SkippedFrames(t *testing.T) {
   `
 	run(cmd)
 }
+
+func audioOnlySegment(t *testing.T, accel Acceleration) {
+	// Reproducing #203
+	run, dir := setupTest(t)
+	defer os.RemoveAll(dir)
+
+	cmd := `
+    # run segmenter and sanity check frame counts . Hardcode for now.
+    ffmpeg -loglevel warning -i "$1"/../transcoder/test.ts -c:a copy -c:v copy -f hls test.m3u8
+    ffprobe -loglevel warning -select_streams v -count_frames -show_streams test0.ts | grep nb_read_frames=120
+    ffprobe -loglevel warning -select_streams v -count_frames -show_streams test1.ts | grep nb_read_frames=120
+    ffprobe -loglevel warning -select_streams v -count_frames -show_streams test2.ts | grep nb_read_frames=120
+    ffprobe -loglevel warning -select_streams v -count_frames -show_streams test3.ts | grep nb_read_frames=120
+
+    # drop all video frames from seg2
+    ffmpeg -loglevel warning -i test2.ts -c:a copy -c:v libx264 -vf "fps=fps=1/100000" test02.ts
+    mv test02.ts test2.ts
+
+    # verify no video frames in seg2
+    ffprobe -loglevel warning -show_streams -select_streams v -count_frames test2.ts | grep nb_read_frames=N/A
+  `
+	run(cmd)
+
+	// Test encoding with audio-only segment in between stream
+	tc := NewTranscoder()
+	prof := P144p30fps16x9
+	for i := 0; i < 4; i++ {
+		in := &TranscodeOptionsIn{
+			Fname: fmt.Sprintf("%s/test%d.ts", dir, i),
+			Accel: accel,
+		}
+		out := []TranscodeOptions{{
+			Oname:   fmt.Sprintf("%s/out%d.ts", dir, i),
+			Profile: prof,
+			Accel:   accel,
+		}}
+		_, err := tc.Transcode(in, out)
+		if err != nil {
+			t.Error(err)
+		}
+	}
+	tc.StopTranscoder()
+
+	// Test encoding with audio-only segment in start of stream
+	tc = NewTranscoder()
+	defer tc.StopTranscoder()
+	for i := 2; i < 4; i++ {
+		in := &TranscodeOptionsIn{
+			Fname: fmt.Sprintf("%s/test%d.ts", dir, i),
+			Accel: accel,
+		}
+		out := []TranscodeOptions{{
+			Oname:   fmt.Sprintf("%s/out2_%d.ts", dir, i),
+			Profile: prof,
+			Accel:   accel,
+		}}
+		_, err := tc.Transcode(in, out)
+		if err != nil {
+			t.Error(err)
+		}
+	}
+}
+
+func TestTranscoder_AudioOnly(t *testing.T) {
+	audioOnlySegment(t, Software)
+}
--- a/ffmpeg/ffmpeg.go
+++ b/ffmpeg/ffmpeg.go
@@ -41,6 +41,7 @@ type ComponentOptions struct {
 type Transcoder struct {
 	handle  *C.struct_transcode_thread
 	stopped bool
+	started bool
 	mu      *sync.Mutex
 }
 
@@ -183,6 +184,16 @@ func (t *Transcoder) Transcode(input *TranscodeOptionsIn, ps []TranscodeOptions)
 	hw_type, err := accelDeviceType(input.Accel)
 	if err != nil {
 		return nil, err
+	}
+	if !t.started {
+		ret := int(C.lpms_is_bypass_needed(C.CString(input.Fname)))
+		if ret != 1 {
+			// no bypass needed, continue
+			t.started = true
+		} else {
+			// audio-only segment, bypassed transcoding
+			return bypassTranscoding(input, ps)
+		}
 	}
 	fname := C.CString(input.Fname)
 	defer C.free(unsafe.Pointer(fname))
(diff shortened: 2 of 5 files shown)


ffmpeg/ffmpeg.go
```diff
@@ -171,6 +172,28 @@ func Transcode3(input *TranscodeOptionsIn, ps []TranscodeOptions) (*TranscodeRes
 	return t.Transcode(input, ps)
 }
 
+// Force copy-transcoding for audio-only segments
+func BypassTranscoding(input *TranscodeOptionsIn, ps []TranscodeOptions) (*TranscodeResults, error) {
+	glog.Errorf("WARNING: Input %s has audio & video streams but no video frames. Bypassing the transcoder.\n", input.Fname)
```
yondonfu, line 177: Seems like `glog.Warningf()` would be a good fit here.

ffmpeg/ffmpeg.go
```diff
@@ -171,6 +172,28 @@ func Transcode3(input *TranscodeOptionsIn, ps []TranscodeOptions) (*TranscodeRes
 	return t.Transcode(input, ps)
 }
 
+// Force copy-transcoding for audio-only segments
+func BypassTranscoding(input *TranscodeOptionsIn, ps []TranscodeOptions) (*TranscodeResults, error) {
```
yondonfu, line 176: Could this function be private? Doesn't seem like it needs to be a part of the public API.

ffmpeg/ffmpeg.go
```diff
@@ -184,6 +207,17 @@ func (t *Transcoder) Transcode(input *TranscodeOptionsIn, ps []TranscodeOptions)
...
 		return nil, err
 	}
+	if !t.started {
+		ret := int(C.lpms_is_bypass_needed(C.CString(input.Fname)))
+		if ret != 1 {
+			// no bypass needed
+			t.started = true
+		} else {
+			// audio-only segment, bypassed transcoding
+			res, err := BypassTranscoding(input, ps)
```
yondonfu, line 217: This could be shortened to:

```golang
return BypassTranscoding(input, ps)
```

ffmpeg/api_test.go
```diff
@@ -1175,3 +1175,69 @@ func TestTranscoder_SkippedFrames(t *testing.T) {
...
+	cmd := `
+    # run segmenter and sanity check frame counts . Hardcode for now.
+    ffmpeg -loglevel warning -i "$1"/../transcoder/test.ts -c:a copy -c:v copy -f hls test.m3u8
+    ffprobe -loglevel warning -select_streams v -count_frames -show_streams test0.ts | grep nb_read_frames=120
+    ffprobe -loglevel warning -select_streams v -count_frames -show_streams test1.ts | grep nb_read_frames=120
+    ffprobe -loglevel warning -select_streams v -count_frames -show_streams test2.ts | grep nb_read_frames=120
+    ffprobe -loglevel warning -select_streams v -count_frames -show_streams test3.ts | grep nb_read_frames=120
+
+    # drop all video frames from seg2
+    ffmpeg -loglevel warning -i test2.ts -c:a copy -c:v libx264 -vf "fps=fps=1/100000" test02.ts
```
yondonfu, line 1193: Question for my own personal understanding:

Do we drop all video frames here by setting the output fps to 1/100000? I'm guessing that in this case the fps is so low that no video frames end up in the output?

ffmpeg/lpms_ffmpeg.c
```diff
@@ -272,6 +272,45 @@ int lpms_rtmp2hls(char *listen, char *outf, char *ts_tmpl, char* seg_time, char
...
+  if (ret < 0) { ret = -1; goto close_format_context; }
+  ret = avformat_find_stream_info(ic, NULL);
+  if (ret < 0) { ret = -1; goto close_format_context; }
+
+  vstream = av_find_best_stream(ic, AVMEDIA_TYPE_VIDEO, -1, -1, NULL, 0);
+  astream = av_find_best_stream(ic, AVMEDIA_TYPE_AUDIO, -1, -1, NULL, 0);
+  if (vstream >= 0 && astream >= 0) {
+      if (AV_PIX_FMT_NONE == ic->streams[vstream]->codecpar->format &&
+          0 == ic->streams[vstream]->codecpar->height) {
+          // no valid pixel format and picture height => needs bypass
```
yondonfu, line 298: Just to clarify: we treat AV_PIX_FMT_NONE as a signal that the video stream has 0 frames? Could there ever be a scenario where there is no valid pixel format, but the video stream has > 0 frames?

ffmpeg/ffmpeg.go
```diff
@@ -171,6 +172,28 @@ func Transcode3(input *TranscodeOptionsIn, ps []TranscodeOptions) (*TranscodeRes
...
+				p.VideoEncoder = ComponentOptions{Name: "copy"}
+			}
+			if "drop" != p.AudioEncoder.Name && "copy" != p.AudioEncoder.Name {
+				p.AudioEncoder = ComponentOptions{Name: "copy"}
+			}
+			ps[i] = p
+		}
+	}
+	// use a new thread to transcode
+	t := NewTranscoder()
```
yondonfu, line 191: What is the advantage of creating a new transcoder in this function as opposed to just overriding the video profiles to use the copy video & audio encoder in `Transcode()`? 

i.e. in `Transcode()`

```golang
if !t.started {
    ret := int(C.lpms_is_bypass_needed(C.CString(input.Fname)))
    if ret != 1 {
        t.started = true
    } else {
        // Override video profiles to use copy video & audio encoder
        // Then, proceed with the rest of the Transcode() method
    }
}
```

iameli:
Right on — testing this now against the MistServer environment where we were encountering the problem.

ffmpeg/ffmpeg.go
jailuthra: Yeah I tried that at first.

The problem happens on a subsequent normal segment in the same stream/session. We initialize the decoder/encoder contexts at the start of the session, so the Cgo module is unable to initialize a `non-copy` segment in a session that was started with a  `copy` segment and fails with [an error here](https://github.com/livepeer/lpms/blob/da8d5c9b1d0fd06db086a102f1281137255c288d/ffmpeg/lpms_ffmpeg.c#L652) while opening the stream. I avoided messing around with that logic for this.

ffmpeg/lpms_ffmpeg.c
jailuthra: > Could there ever be a scenario where there is no valid pixel format, but the video stream has > 0 frames?
Yep it might happen for some weird source pixel format that FFmpeg doesn't recognize, but anyway the encoders won't recognize it either so it'll fail at a later stage unless the user selected `copy`.

But with the extra check of checking `codecpar->height` ([AVCodecParameters.height](https://ffmpeg.org/doxygen/trunk/structAVCodecParameters.html#a1ec57ee84f19cf65d00eaa4d2a2253ce)) here is a reliable indicator of having 0-frames, even more so than `streavs[vstream]->nb_frames` ([AVStream.nb_frames](https://ffmpeg.org/doxygen/trunk/structAVStream.html#a4382c3064df1c9eb232ac198dec067f9)) which is not always set by default.

ffmpeg/api_test.go
jailuthra: > Do we drop all video frames here by setting the output fps to 1/100000? I'm guessing that in this case the fps is so low that no video frames end up in the output?

Yes exactly! I first used the `-vn` parameter instead of FPS filter, but that dropped the video stream altogether, which did not match the scenario in the MistServer samples

ffmpeg/ffmpeg.go
jailuthra: Ah dang I keep forgetting capitalization => scope in golang, thanks!

ffmpeg/ffmpeg.go
jailuthra: Ah good catch will fix!

ffmpeg/ffmpeg.go
jailuthra: > Seems like glog.Warningf() would be a good fit here.
Ah had no idea this existed, perfect!

> Side note: Do we want the extra newline at the end? Just wondering if it was purposefully added for visual reasons.
Bad habit from C's printf function, will remove thx! I see glog will add a newline if not present, and leave it as-is if present https://godoc.org/github.com/golang/glog#Warningf

jailuthra:
> Perhaps some of these things could be useful in this repo as well.
Ah TIL, I had no idea about fixup/squashing in github, will use it from now thanks!

PR #206, opened 2020-08-25
ffmpeg: bypass txcoder only for mpegts

jailuthra:
**Why?**
Refer to https://github.com/livepeer/go-livepeer/pull/1608 for details on why this was needed.

**What?**
Ensure the bypass implemented for audio-only segments at the start of a stream only happen for MPEG-TS muxer format and help out the transcoder by passing the out format through `VideoProfile.Format`. (go-livepeer uses `.tempfile` extension because the segs might be .ts or .m4s)

**Note:** This is a temporary hack, and I want to further emphasize that the transcoder should support a Go function (that calls Cgo ffmpeg api) to *only check* whether a bypass is needed. The actual bypass logic (and what options to tweak) should be maintained upstream either in the Broadcaster code (if needed for public network) or the API product code.

### DIFF
--- a/ffmpeg/ffmpeg.go
+++ b/ffmpeg/ffmpeg.go
@@ -381,8 +381,11 @@ func (t *Transcoder) Transcode(input *TranscodeOptionsIn, ps []TranscodeOptions)
 func bypassTranscoding(input *TranscodeOptionsIn, ps []TranscodeOptions) (*TranscodeResults, error) {
 	glog.Warningf("Input %s has audio & video streams but no video frames. Bypassing the transcoder.", input.Fname)
 	for i, p := range ps {
-		if p.Profile != (VideoProfile{}) {
-			p.Profile = (VideoProfile{})
+		if p.Profile != (VideoProfile{}) && p.Profile.Format != FormatMP4 {
+			// we only bypass when a VideoProfile is set (i.e. not a copy/drop) and output format
+			// is mpegts.
+			// FIXME: Temporary fix. The transcoder is not the place for such hacky code
+			p.Profile = VideoProfile{Name: p.Profile.Name, Format: FormatMPEGTS}
 			if "drop" != p.VideoEncoder.Name && "copy" != p.VideoEncoder.Name {
 				p.VideoEncoder = ComponentOptions{Name: "copy"}
 			}


yondonfu:
> Ensure the bypass implemented for audio-only segments at the start of a stream only happen for MPEG-TS muxer format

Does this mean that the transcoder won't be able handle successive MP4 segments if the first one is audio-only?

jailuthra:
> Does this mean that the transcoder won't be able handle successive MP4 segments if the first one is audio-only?

Yep, deep dived into it further before commenting here, and turns out the MP4 format doesn't support dimensionless (0-frame) video streams at all. So there doesn't seem to be a way to copy an mpeg-ts segment with the 0-frame video stream, to an mp4 segment while still retaining the stream (it by default copies it to an audio-stream only segment, which doesn't work well with any player treating the segments as part of same stream)

The current behaviour (as of this PR) of the transcoder with MP4 output is to skip the audio-only segments all together. It remains the same even if I allow MP4 segments to go thorugh the `bypassTranscoder` function as it fails later on while opening output.

The only way that we can support this in the long-term is creating a dummy video stream and using that whenever the user supplied video stream is broken. That seemed simple to me initially but that will also have problems in case the user's source video is not the aspect ratio selected in the `VideoProfile` struct as then we'll have no way to know the input aspect ratio until we actually see a valid frame. Will think about this more!

ffmpeg/ffmpeg.go
```diff
@@ -381,8 +381,11 @@ func (t *Transcoder) Transcode(input *TranscodeOptionsIn, ps []TranscodeOptions)
 func bypassTranscoding(input *TranscodeOptionsIn, ps []TranscodeOptions) (*TranscodeResults, error) {
 	glog.Warningf("Input %s has audio & video streams but no video frames. Bypassing the transcoder.", input.Fname)
 	for i, p := range ps {
-		if p.Profile != (VideoProfile{}) {
-			p.Profile = (VideoProfile{})
+		if p.Profile != (VideoProfile{}) && p.Profile.Format != FormatMP4 {
+			// we only bypass when a VideoProfile is set (i.e. not a copy/drop) and output format
+			// is mpegts.
+			// FIXME: Temporary fix. The transcoder is not the place for such hacky code
+			p.Profile = VideoProfile{Name: p.Profile.Name, Format: FormatMPEGTS}
```
yondonfu, line 388: We'll skip the non-encoding related operations such as the scale & FPS filter if we initialize a new video profile here and only set the name/format. Is this desired behavior?

My initial thought is that we would want to still apply filters configured in the video profile fields, but just use the copy video & audio encoder. Or is it the case that the filters would not work properly for the first audio-only segment in a session so the only thing we can do is to pass the segment to the copy video & audio encoder?

jailuthra: > or is it the case that the filters would not work properly for the first audio-only segment in a session so the only thing we can do is to pass the segment to the copy video & audio encoder?

Yeah exactly! Filters get applied on decoded frames, and there are no video frames here for the audio-only segments :)
Although the subsequent normal segments will not go through this bypass function, and would get filtered/transcoded as usual.

There's no way to even set the filters on in the `copy` mode https://stackoverflow.com/questions/53518589/how-to-use-filtering-and-stream-copy-together-with-ffmpeg

We can try enabling the audio-filters here (which don't really do much) but I wanna avoid that at first and only consider it if we face issues, because doing that will likely complicate the delicate re-muxing further.

yondonfu (approved):
LGTM

PR #207, opened 2020-09-24
ffmpeg: Throw error instead of bypassing for audio-only stream

jailuthra:
**What is done?**
* Reverted the bypassing performed for initial segments with broken video streams.
* Retained the detection code and now throw an early error for such streams to prevent the extra-cost associated with initializing contexts for decoding.

**Why?**
The original error is not reproducible anymore, and the current state of `master` which bypasses segments altogether may cause issues with playback and is not a stable well-tested solution.


yondonfu (approved):
LGTM

PR #210, opened 2020-10-16
ffmpeg: Fix cstring memleak

jailuthra:
Fix a newbie Cgo error of not freeing `C.CString` that golang's GC doesn't know about.


yondonfu (approved):
LGTM 🚢  

Nice catch!

PR #212, opened 2020-11-09
Re-organizing LPMS Cgo codebase

jailuthra:
**Why**

All the C functions managing FFmpeg library structs and other internal state variables for the transcoding and segmenter pipelines were a part of a single Cgo module `lpms_ffmpeg.c`. This was a bit hard to maintain and things felt all over the place.

**What**

Split the single Cgo module into the following:

- **extras.c/.h** - Functions unrelated to main transcode pipeline like segmenting and stream validation
- **transcoder.c/.h** - Main transcode loop and other setup functions

The transcoder module further divides into sub-modules to handle specific phases of the pipeline:

- **decoder.c/.h** - Input-to-raw-frames i.e demuxing, decoding
- **filter.c/.h** - Filtering raw frames i.e. scaling, FPS changes etc
- **encoder.c/.h** - Raw-frames-to-output i.e. encoding, muxing

Earlier the `process_out` method was very long and had a lot of filtering and encoding code mixed up. In this PR I've moved the filtering specific code from `encoder.c -> process_out()` to `filter.c -> filtergraph_write()/filtergraph_read()` which handle sending and receiving frames from FFmpeg filtergraph and do the timestamp magic.

No functional changes in this PR, thus no unit tests have been broken and all functionality works like before.


jailuthra:
The `transcoder.c` module can further be cleaned up as it has some open input/output functions that can better be placed in one of the submodules. But I held off on doing that for now as those can take a bit more improvement in their content and can be moved when that's done.

oscar-davids:
Regarding lpms error handling

I believe that the above work conforms to the code design principle (SOLID) and is necessary work for the future expansion of additional services. I think it would be better to handle the following in the whole code.

1. redefined enum type error values

Although it is old code, redefining the error values already defined as int type in ffmpeg to enum type again in lpms can lead to confusion in code debugging in the future. How about using the prefix LPMS instead?

AVERROR_BSF_NOT_FOUND -> LPMSERROR_BSF_NOT_FOUND

https://github.com/FFmpeg/FFmpeg/blob/master/libavutil/error.h#L49

https://github.com/livepeer/lpms/blob/8001f9043b1059224c23adeb91d7948364e7dfac/ffmpeg/ffmpeg_errors.h#L8

2. redefinition lpms error processing parts

    There are many error definitions in the current code, for example #define main_err, #define proc_err, #define, #undef. 
    I guess It is more than one hundred lines.
    I think that batching this part in ffmpeg_errors.h can reduce the amount of code and increase the readability.

> #define lpms_err(msg, label) {\
>   char errstr[AV_ERROR_MAX_STRING_SIZE] = {0}; \
>   if (!ret) ret = AVERROR(EINVAL); \
>   if (ret <-1) av_strerror(ret, errstr, sizeof errstr); \
>   fprintf(stderr, "%s: %s\n", msg, errstr); \
>   goto label; \
> }

or 

> #define lpms_err( msg, errval, label) {\
>   char errstr[AV_ERROR_MAX_STRING_SIZE] = {0}; \
>   if (!ret) ret = AVERROR(errval); \
>   if (ret <-1) av_strerror(ret, errstr, sizeof errstr); \
>   fprintf(stderr, "%s: %s: %s\n", __FUNCTION\__, msg, errstr); \
>   goto label; \
> }

When using it, we can use it as follows.  

> if (ret <0) lpms_err("Unable to reopen demuxer", transcode_cleanup);

and I think It would be nice to put "av_strerror" code in this part for debugging.

https://github.com/livepeer/lpms/blob/8001f9043b1059224c23adeb91d7948364e7dfac/ffmpeg/encoder.c#L257

jailuthra:
> There are many error definitions in the current code, for example #define main_err, #define proc_err, #define, #undef.
I guess It is more than one hundred lines.
I think that batching this part in ffmpeg_errors.h can reduce the amount of code and increase the readability.

@oscar-davids That's a great idea, will do that! I think it would be better suited to the other PR (to be applied on top of this one) I have open (#214) around improving the logging.

Feel free to review that PR meanwhile too in case you have other suggestions around the logging code!

oscar-davids (approved):

PR #214, opened 2020-11-20
ffmpeg: Improve Cgo logging

### LINKED ISSUE #209: Improve Logging
Currently the logs outputted by LPMS are a weird mix of print statements from Go, Cgo and Ffmpeg library output on stderr.
This makes it hard to debug and the overall node logs quite messy.

It would really benefit visibility if we have a standardized log format across the languages/libraries for `stderr` and also probably have verbose ffmpeg library logs into a separate file (because we can't catch and re-emit logs for every case) instead of mixing it with rest of the logs.

jailuthra:
**Why**
Clean up logging in the Cgo code (Fixes #209)

**What**
- Use ffmpeg's logging library with appropriate levels like ERROR, WARNING, INFO instead of throwing everything on stderr blindly.
- Add a well defined log prefix like `[lpms.decoder.xyz]` of the component from where the log is being emitted.
- Does *not* add any extra INFO logs, will do it separately the next time I go around debugging something and need to trace out the location in the transcoder pipeline

**Notes**
Commit(s) to be applied on top of #212

### DIFF
--- a/ffmpeg/decoder.c
+++ b/ffmpeg/decoder.c
@@ -1,5 +1,6 @@
 #include "transcoder.h"
 #include "decoder.h"
+#include "logging.h"
 
 #include <libavutil/pixfmt.h>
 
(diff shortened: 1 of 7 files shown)


oscar-davids:
I think it's a must for code maintenance and debugging.

As we know, ffmpeg logging is leveled by the "av_log_set_level" function. But this function has been hard-coded as "AV_LOG_WARNING" in our C level lpms_init() function.
I think it would be better to redefine the C level lpms_init() function and the go level InitFFmpeg() function for outputting the log.

> type LogLevel int
> 
> const (
> 	LogPanic LogLevel = iota
> 	LogFatal
> 	LogError
> 	LogWarning
> 	LogInfo
> 	LogVerbose
> 	LogDebug
> 	LogTrace
> )	
> 
> func InitFFmpeg(level LogLevel ) {
>     ...
> }

Along with this, I think that if some stuff discussed in #212 would be supplemented, it would be a better development environment.

jailuthra:
> But this function has been hard-coded as "AV_LOG_WARNING" in our C level lpms_init() function.

@oscar-davids Ah yes I had thought about making that match the golang's glog logging level but left it as-is because matching the levels might lead to really verbose ffmpeg logs.

But the suggestion you make around making it separately configurable using a golang enum sounds like a good way to go! Will do this in the update commit.

> I think it would be better to redefine the C level lpms_init() function and the go level InitFFmpeg() function for outputting the log.

jailuthra:
Updates:
* Create shared `#define` macros for the logging functions for ERROR, WARNING and INFO levels
* Make logging level configurable via the C and Golang module's init() methods

oscar-davids (approved):

PR #220, opened 2021-01-20
decoder: Error out on segments with no keyframe packets

### LINKED ISSUE #219: Transcoder stuck for input segments with no keyframes but valid stream information
Forking this from https://github.com/livepeer/go-livepeer/issues/1725#issuecomment-763074766

We are currently able to handle no-video-frame segments after fixing #203, but that check relies on those segments not having valid video metadata like resolution or bitrate.

Recently we are encountering a new edge-case mpegts segment that has no video frames but valid metadata (like one of the segments in [this tar](https://storage.googleapis.com/docs.livepeer.live/eli/segment-debugging/kryptonite.tar.gz)) causing the transcoder session to get stuck.

UPDATE: Looks like that particular segment does contain valid encoded data but none of the frames are marked as I-Frame or keyframe

**Root Cause:**

Most likely this happens because of the decoder flush logic, which currently relies on sending the first keyframe packet to flush out frames from the decoder, but would keep looping if no such packet found.

https://github.com/livepeer/lpms/blob/6cbff98801791f3b43b08a888f3b3302fb44c3c1/ffmpeg/decoder.c#L91-L103

**Likely Fix:**

Update the decoder flush logic to handle this case.

jailuthra:
**Why?**
The video decoder currently gets stuck in an infinite loop while flushing if the input segment does not contain any keyframes, as it uses the first keyframe packet to flush the decoder buffer.

**What?**
If no keyframe packet is found after reading and sending all packets to the decoder, return an explicit error and don't try to flush.

**Testing**
- [x] Manually tested on a segment with no keyframes
- [x] Existing unit test suite works
- [x] Added a commented out test case for segment with no keyframes

**Fixes**
Fixes #219

### DIFF
--- a/ffmpeg/api_test.go
+++ b/ffmpeg/api_test.go
@@ -1243,3 +1243,46 @@ func audioOnlySegment(t *testing.T, accel Acceleration) {
 func TestTranscoder_AudioOnly(t *testing.T) {
 	audioOnlySegment(t, Software)
 }
+
+/*
+func noKeyframeSegment(t *testing.T, accel Acceleration) {
+	// Reproducing #219
+	run, dir := setupTest(t)
+	defer os.RemoveAll(dir)
+
+	cmd := `
+		cp "$1"/../transcoder/kryp-*.ts .
+
+    # verify no keyframes in kryp-2.ts but there in kryp-1.ts
+    ffprobe -select_streams v -show_streams -show_packets kryp-1.ts | grep flags=K | wc -l | grep 1
+    ffprobe -select_streams v -show_streams -show_packets kryp-2.ts | grep flags=K | wc -l | grep 0
+  `
+	run(cmd)
+
+	tc := NewTranscoder()
+	prof := P144p30fps16x9
+	for i := 1; i <= 2; i++ {
+		in := &TranscodeOptionsIn{
+			Fname: fmt.Sprintf("%s/kryp-%d.ts", dir, i),
+			Accel: accel,
+		}
+		out := []TranscodeOptions{{
+			Oname:   fmt.Sprintf("%s/out%d.ts", dir, i),
+			Profile: prof,
+			Accel:   accel,
+		}}
+		_, err := tc.Transcode(in, out)
+		if i == 2 && (err == nil || err.Error() != "No keyframes in input") {
+			t.Error("Expected to fail for no keyframe segment but did not")
+		} else if i != 2 && err != nil {
+			t.Error(err)
+		}
+	}
+	tc.StopTranscoder()
+
+}
+
+func TestTranscoder_NoKeyframe(t *testing.T) {
+	noKeyframeSegment(t, Software)
+}
+*/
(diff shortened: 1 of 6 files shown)


ffmpeg/transcoder.c
```diff
@@ -193,7 +194,9 @@ int transcode(struct transcode_thread *h,
     if (ret == AVERROR_EOF) break;
                             // Bail out on streams that appear to be broken
     else if (lpms_ERR_PACKET_ONLY == ret) ; // keep going for stream copy
-    else if (ret < 0) LPMS_ERR(transcode_cleanup, "Could not decode; stopping");
+    else if (lpms_ERR_INPUT_NOKF == ret) {
+      LPMS_ERR(transcode_cleanup, "Could not decode; No keyframes found in the input");
```
yondonfu, line 198: Nit: Should the second part of this error string just exactly match the description for `lpms_ERR_INPUT_NOKF` for consistency?

ffmpeg/api_test.go
```diff
@@ -1212,7 +1212,9 @@ func audioOnlySegment(t *testing.T, accel Acceleration) {
 			Accel:   accel,
 		}}
 		_, err := tc.Transcode(in, out)
-		if err != nil {
+		if i == 2 && (err == nil || err.Error() != "No keyframes in input") {
+			t.Error("Expected to fail for audio-only segment midstream but did not")
```
yondonfu, line 1216: Just to clarify - at the moment, the transcoder is able to handle audio-only segments mid-stream (we currently only reject audio-only segments at the beginning of a stream), but after this PR the transcoder will no longer be able to handle audio-only segments because they don't have any keyframes? 

Should the transcoder instead continue accepting audio-only segments mid-stream, but only error out on > 0 frame segments with no keyframes? Perhaps by returning `lpms_ERR_INPUT_NOKF` from the cgo code only if no keyframes were detected *and* at least 1 frame was fed to the decoder?

jailuthra: Yep this is an intentional stop-gap decision to treat both the 0-video-packets and 0-keyframe-video-packets cases the same way.

A con of this is, as you mention, we no longer support 0-video-packets segments mid-stream even though they could be supported. IIRC my testing of that particular usecase anyway did not work well with some end-user HLS players as their player's "decoder" also had similar issues around non-existent video metadata midstream. So if we don't send that segment back to the user's player altogether, things might actually be smoother.

A pro of this stop-gap is that we don't need to separately create a new testcase for both, and it keeps the underlying code clean and doesn't introduce a new "frames decoded" counter, which we can reconsider and add later on.

yondonfu: > IIRC my testing of that particular usecase anyway did not work well with some end-user HLS players as their player's "decoder" also had similar issues around non-existent video metadata midstream.

I thought this was just for audio-only segments at the beginning of a stream which is why the transcoder will return an error immediately for audio-only segments at the beginning of a stream? Reference #207.

If the transcoder does handle audio-only segments mid-stream smoothly right now, I think we should avoid changing that behavior here if possible.

jailuthra: Done, also turns out I already had a "frames sent" counter called `pkt_diff` from earlier, so was able to re-use it ezpz :)

yondonfu (approved):
LGTM

PR #223, opened 2021-02-03
decoder: update sentinel counter even for bad packets

### LINKED ISSUE #222: Stuck in error loop if decoding fails for flush packet
From @yondonfu 
```
ERROR: decoder.c:32] Error sending flush packet : Generic error in an external library
Feb 01 07:52:48 ber-prod-livepeer-orchestrator-2 orchestrator: [h264_cuvid @ 0x7f44d0122480] ctx->cvdl->cuvidParseVideoData(ctx->cuparser, &cupkt) failed -> CUDA_ERROR_UNKNOWN: unknown error
```

Failure to update the `sentinel_count` here might lead to a stuck loop https://github.com/livepeer/lpms/blob/c9bb34142477475e86f991274a5ab596f50c1a09/ffmpeg/decoder.c#L31-L33

jailuthra:
**What?**
Fix the incrementing logic of the flush "sentinel" packet counter. Previously that counter was not updated if FFmpeg threw an error for the packet, which lead to an infinite loop.

**Why?**
Fixes #222

### DIFF
--- a/ffmpeg/decoder.c
+++ b/ffmpeg/decoder.c
@@ -25,12 +25,13 @@ static int lpms_receive_frame(struct input_ctx *ictx, AVCodecContext *dec, AVFra
 static int send_first_pkt(struct input_ctx *ictx)
 {
   if (ictx->flushed) return 0;
-  if (!ictx->first_pkt) return -1;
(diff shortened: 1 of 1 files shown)


yondonfu (approved):
LGTM

PR #225, opened 2021-04-01
ffmpeg: separate non-retryable errors

jailuthra:
**Why?**
Needed for https://github.com/livepeer/go-livepeer/issues/1774

**What?**
Add more errors, and separate out the "Non Retryable" category to be used by the B. Updated list of errors includes:

- All LPMS C-module errors
- A subset of LPMS Go-module errors around bad input/parameters
- A subset of [internal ffmpeg errors](https://ffmpeg.org/doxygen/trunk/error_8c_source.html#l00034) that may get surfaced back to the B

### DIFF
--- a/ffmpeg/ffmpeg_errors.go
+++ b/ffmpeg/ffmpeg_errors.go
@@ -10,7 +10,7 @@ import (
 	"unsafe"
 )
 
-var LPMSErrors = []struct {
+var lpmsErrors = []struct {
 	Code C.int
 	Desc string
 }{
@@ -40,7 +40,7 @@ func error_map() map[int]error {
 	}
 
 	// Add in LPMS specific errors
-	for _, v := range LPMSErrors {
+	for _, v := range lpmsErrors {
 		m[int(v.Code)] = errors.New(v.Desc)
 	}
 
@@ -48,6 +48,34 @@ func error_map() map[int]error {
 }
 
 var ErrorMap = error_map()
+
+func non_retryable_errs() []string {
+	errs := []string{}
+	// Add in Cgo LPMS specific errors
+	for _, v := range lpmsErrors {
+		errs = append(errs, v.Desc)
+	}
+	// Add in internal FFmpeg errors
+	// from https://ffmpeg.org/doxygen/trunk/error_8c_source.html#l00034
+	ffmpegErrors := []string{
+		"Decoder not found", "Demuxer not found", "Encoder not found",
+		"Muxer not found", "Option not found", "Invalid argument",
+	}
+	for _, v := range ffmpegErrors {
+		errs = append(errs, v)
+	}
+	// Add in ffmpeg.go transcoder specific errors
+	transcoderErrors := []error{
+		ErrTranscoderRes, ErrTranscoderVid, ErrTranscoderFmt,
+		ErrTranscoderPrf, ErrTranscoderGOP, ErrTranscoderDev,
+	}
+	for _, v := range transcoderErrors {
+		errs = append(errs, v.Error())
+	}
+	return errs
+}
+
+var NonRetryableErrs = non_retryable_errs()
 
 // Use of this source code is governed by a MIT license that can be found in the LICENSE file.
 // Corbatto (luca@corbatto.de)
--- a/ffmpeg/api_test.go
+++ b/ffmpeg/api_test.go
@@ -1232,8 +1232,8 @@ func audioOnlySegment(t *testing.T, accel Acceleration) {
 			Accel:   accel,
 		}}
 		_, err := tc.Transcode(in, out)
-		if i == 2 && (err == nil || err.Error() != "No video parameters found while initializing stream") {
-			t.Error("Expected to fail for audio-only segment but did not")
+		if i == 2 && (err == nil || err.Error() != "TranscoderInvalidVideo") {
+			t.Errorf("Expected to fail for audio-only segment but did not, instead got err=%v", err)
 		} else if i != 2 && err != nil {
 			t.Error(err)
 		}
--- a/ffmpeg/ffmpeg.go
+++ b/ffmpeg/ffmpeg.go
@@ -21,10 +21,12 @@ import "C"
 var ErrTranscoderRes = errors.New("TranscoderInvalidResolution")
 var ErrTranscoderHw = errors.New("TranscoderInvalidHardware")
 var ErrTranscoderInp = errors.New("TranscoderInvalidInput")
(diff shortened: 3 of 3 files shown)


oscar-davids:
I think "ErrTranscoderInp" is an ambiguous variable. [Here ](https://github.com/livepeer/lpms/blob/589071faa9bf95a85bbcbdb5b7054422ff10b656/ffmpeg/ffmpeg.go#L184)we can consider this variable as OOM.
How about adding one error variable for device id?
ex. var ErrTranscoderSid = errors.New("TranscoderInvalidSameDeviceID")
And then we can add this variable [here](https://github.com/livepeer/lpms/blob/589071faa9bf95a85bbcbdb5b7054422ff10b656/ffmpeg/ffmpeg_errors.go#L69) and [here](https://github.com/livepeer/lpms/blob/589071faa9bf95a85bbcbdb5b7054422ff10b656/ffmpeg/ffmpeg.go#L148) instead of "ErrTranscoderInp".

ffmpeg/ffmpeg_errors.go
```diff
@@ -49,6 +49,34 @@ func error_map() map[int]error {
...
+func non_retriable_errs() []string {
+	errs := make([]string, 0, 20)
+	// Add in Cgo LPMS specific errors
+	for _, v := range lpmsErrors {
+		errs = append(errs, v.Desc)
+	}
+	// Add in internal FFmpeg errors
+	// from https://ffmpeg.org/doxygen/trunk/error_8c_source.html#l00034
+	ffmpegErrors := []string{
+		"Decoder not found", "Demuxer not found", "Encoder not found",
```
yondonfu, line 61: Curious what the criteria for selecting these specific errors was since I see that there are a lot more documented in the above link?

ffmpeg/ffmpeg_errors.go
```diff
@@ -49,6 +49,34 @@ func error_map() map[int]error {
...
+		ErrTranscoderRes, ErrTranscoderInp, ErrTranscoderVid,
+		ErrTranscoderFmt, ErrTranscoderPrf, ErrTranscoderGOP,
+	}
+	for _, v := range transcoderErrors {
+		errs = append(errs, v.Error())
+	}
+	return errs
+}
+
+var NonRetriableErrs = non_retriable_errs()
```
yondonfu, line 78: Pretty minor spelling comment:

So...it looks like "retryable" and "retriable" are both acceptable spellings at least according to this [SO post](https://english.stackexchange.com/questions/305273/retriable-or-retryable/305433) (apparently neither is in most dictionaries). But, we already use the [retryable spelling](https://github.com/livepeer/go-livepeer/search?q=retryable) in a bunch of places in go-livepeer so for consistency I recommend using that spelling.

ffmpeg/ffmpeg_errors.go
```diff
@@ -49,6 +49,34 @@ func error_map() map[int]error {
 
 var ErrorMap = error_map()
 
+func non_retriable_errs() []string {
+	errs := make([]string, 0, 20)
```
yondonfu, line 53: Maybe we can use a nil slice that is appended to so we don't have to worry about a hardcoded capacity? Incurs additional allocations under the hood, but probably ok since this function is called once to set the exported package variable.

ffmpeg/ffmpeg_errors.go
jailuthra: No specific criteria - picked the few that immediately seemed related to bad input/parameters. I've personally only seen "Invalid argument" often.

Plan was to iterate upon this list if we see particular internal errors propping up in future that can be classified as retriable. WDYT?

ffmpeg/ffmpeg_errors.go
jailuthra: Ah good catch, I used them interchangeably without even noticing. Will fix!

ffmpeg/ffmpeg_errors.go
```diff
@@ -49,6 +49,34 @@ func error_map() map[int]error {
...
+	ffmpegErrors := []string{
+		"Decoder not found", "Demuxer not found", "Encoder not found",
+		"Muxer not found", "Option not found", "Invalid argument",
+	}
+	for _, v := range ffmpegErrors {
+		errs = append(errs, v)
+	}
+	// Add in ffmpeg.go transcoder specific errors
+	transcoderErrors := []error{
+		ErrTranscoderRes, ErrTranscoderInp, ErrTranscoderVid,
```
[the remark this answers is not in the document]
jailuthra: I didn't add `ErrTranscoderStp` there precisely because of that. Is this a typo, or am I missing some context?

jailuthra:
>  How about adding one error variable for device id?
@oscar-davids Good idea, I've added a new `ErrTranscoderDev` error in the fixup commit.

ffmpeg/ffmpeg_errors.go
```diff
@@ -49,6 +49,34 @@ func error_map() map[int]error {
...
+	ffmpegErrors := []string{
+		"Decoder not found", "Demuxer not found", "Encoder not found",
+		"Muxer not found", "Option not found", "Invalid argument",
+	}
+	for _, v := range ffmpegErrors {
+		errs = append(errs, v)
+	}
+	// Add in ffmpeg.go transcoder specific errors
+	transcoderErrors := []error{
+		ErrTranscoderRes, ErrTranscoderInp, ErrTranscoderVid, ErrTranscoderFmt,
```
oscar-davids, line 69: I think "ErrTranscoderInp" is OOM variable, what do you think of this?

jailuthra: I see that after this PR the `ErrTranscoderInp` error is only used for the case when the function is called with `nil` input. https://github.com/livepeer/lpms/blob/560b4f72440a16cc08464608f27767bb9f171b32/ffmpeg/ffmpeg.go#L185

Can you elaborate a bit on what is meant by OOM here?

jailuthra: > I see that after this PR the ErrTranscoderInp error is only used for the case when the function is called with nil input. 

In any case, this error will no longer be triggered by a 'bad video' segment, and only in case somebody uses the `ffmpeg.Transcode(` api with nil input. Will remove it from this list in an update!

ffmpeg/ffmpeg_errors.go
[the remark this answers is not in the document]
yondonfu: Oops I misread ErrTranscoderInp as ErrTranscoderStp...

yondonfu (approved):
LGTM

PR #228, opened 2021-04-29
Scene classification prototype

oscar-davids:
**Motivation**

Since the time the lpmsdemo fork was created, the LPMS codebase architecture has changed quite a bit. 
The ideal architecture is to implement with inference DNN filter, not integrating directly into LPMS C code.
So the source code was modified to reflect this idea.
Following figure shows the workflow.

image

**Build & Test**

-FFmpeg build

To use lvpdnn inference filter, clone from [here ](https://github.com/oscar-davids/FFmpeg/tree/dev-lpvdnn)and must be built with -enable-libtensorflow

-Test

making or getting a trained model, refer to [ModelPath](https://github.com/oscar-davids/lpms/blob/3ebb3493fb6b4a1111ac9d0e6cb3ebdeb204db3e/ffmpeg/videoprofile.go#L97)

software processing

go run cmd/transcoding/transcoding.go transcoder/test.ts **PDnnDetector**,P240p30fps16x9 sw

nvidia processing, GPU number 2

go run cmd/transcoding/transcoding.go transcoder/test.ts **PDnnDetector**,P240p30fps16x9 nv 2

**Key Index**

USE_LVPDNN_ : macro for  lpms C

### DIFF
--- a/cmd/scenedetection/scenedetection.go
+++ b/cmd/scenedetection/scenedetection.go
@@ -18,8 +18,9 @@ func validRenditions() []string {
 }
 
 func main() {
-	if len(os.Args) <= 3 {
-		panic("Usage: <input file> <output renditions, comma separated> <sw/nv>")
+	if len(os.Args) <= 4 {
+		//0,1 input.mp4 P720p25fps16x9,P720p30fps4x3 nv 0
+		panic("Usage:<dnn init deviceids,comma separated> <input file> <output renditions, comma separated> <sw/nv>")
 	}
 	str2accel := func(inp string) (ffmpeg.Acceleration, string) {
 		if inp == "nv" {
@@ -39,9 +40,10 @@ func main() {
 		}
 		return profs
 	}
-	fname := os.Args[1]
-	profiles := str2profs(os.Args[2])
-	accel, lbl := str2accel(os.Args[3])
+	deviceids := os.Args[1]
+	fname := os.Args[2]
+	profiles := str2profs(os.Args[3])
+	accel, lbl := str2accel(os.Args[4])
 
 	profs2opts := func(profs []ffmpeg.VideoProfile) []ffmpeg.TranscodeOptions {
 		opts := []ffmpeg.TranscodeOptions{}
@@ -49,27 +51,41 @@ func main() {
 			o := ffmpeg.TranscodeOptions{
 				Oname:   fmt.Sprintf("out_%s_%d_out.mkv", lbl, i),
 				Profile: profs[i],
-				// Uncomment the following to test scene classifier
-				// Detector: &ffmpeg.DSceneAdultSoccer,
-				Accel: accel,
+				Accel:   accel,
 			}
 			opts = append(opts, o)
 		}
+		//add detection profile
+		o := ffmpeg.TranscodeOptions{
+			Oname:    "out_null.mkv",
+			Profile:  ffmpeg.P144p30fps16x9,
+			Detector: &ffmpeg.DSceneAdultSoccer,
+			Accel:    accel,
+		}
+		opts = append(opts, o)
 		return opts
 	}
 	options := profs2opts(profiles)
 
 	var dev string
 	if accel == ffmpeg.Nvidia {
-		if len(os.Args) <= 4 {
+		if len(os.Args) <= 5 {
 			panic("Expected device number")
 		}
-		dev = os.Args[4]
+		dev = os.Args[5]
 	}
 
-	ffmpeg.InitFFmpeg()
-
 	t := time.Now()
+	err := ffmpeg.InitFFmpegWithDetectorProfile(&ffmpeg.DSceneAdultSoccer, deviceids)
+	defer ffmpeg.ReleaseFFmpegDetectorProfile()
+	end := time.Now()
+
+	if err != nil {
+		panic("Could not initializ DNN engine!")
+	}
+	fmt.Printf("InitFFmpegWithDetectorProfile time %0.4v\n", end.Sub(t).Seconds())
+
+	t = time.Now()
 	fmt.Printf("Setting fname %s encoding %d renditions with %v\n", fname, len(options), lbl)
 	res, err := ffmpeg.Transcode3(&ffmpeg.TranscodeOptionsIn{
 		Fname:  fname,
@@ -79,13 +95,13 @@ func main() {
 	if err != nil {
 		panic(err)
 	}
-	end := time.Now()
+	end = time.Now()
 	fmt.Printf("profile=input frames=%v pixels=%v\n", res.Decoded.Frames, res.Decoded.Pixels)
 	for i, r := range res.Encoded {
 		if r.DetectData != nil {
-			fmt.Printf("profile=%v frames=%v pixels=%v detectdata=%v\n", profiles[i].Name, r.Frames, r.Pixels, r.DetectData)
+			fmt.Printf("profile=%v frames=%v pixels=%v detectdata= %v\n", options[i].Profile, r.Frames, r.Pixels, r.DetectData)
 		} else {
-			fmt.Printf("profile=%v frames=%v pixels=%v\n", profiles[i].Name, r.Frames, r.Pixels)
+			fmt.Printf("profile=%v frames=%v pixels=%v\n", options[i].Profile, r.Frames, r.Pixels)
 		}
 	}
 	fmt.Printf("Transcoding time %0.4v\n", end.Sub(t).Seconds())
--- a/ffmpeg/api_test.go
+++ b/ffmpeg/api_test.go
@@ -1244,8 +1244,6 @@ func TestTranscoder_AudioOnly(t *testing.T) {
 	audioOnlySegment(t, Software)
 }
 
-<<<<<<< HEAD
-<<<<<<< HEAD
 func outputFPS(t *testing.T, accel Acceleration) {
 	run, dir := setupTest(t)
 	defer os.RemoveAll(dir)
@@ -1295,8 +1293,6 @@ func TestTranscoder_OutputFPS(t *testing.T) {
 	outputFPS(t, Software)
 }
 
-=======
->>>>>>> ffmpeg: Comment out detection related tests
 /*
 func detectionFreq(t *testing.T, accel Acceleration) {
 	run, dir := setupTest(t)
--- a/ffmpeg/api_test.go
+++ b/ffmpeg/api_test.go
@@ -1244,6 +1244,8 @@ func TestTranscoder_AudioOnly(t *testing.T) {
 	audioOnlySegment(t, Software)
 }
 
+<<<<<<< HEAD
+<<<<<<< HEAD
 func outputFPS(t *testing.T, accel Acceleration) {
 	run, dir := setupTest(t)
 	defer os.RemoveAll(dir)
@@ -1292,6 +1294,64 @@ func outputFPS(t *testing.T, accel Acceleration) {
 func TestTranscoder_OutputFPS(t *testing.T) {
 	outputFPS(t, Software)
 }
+
+=======
+>>>>>>> ffmpeg: Comment out detection related tests
+/*
+func detectionFreq(t *testing.T, accel Acceleration) {
+	run, dir := setupTest(t)
+	defer os.RemoveAll(dir)
+	cmd := `
+    # run segmenter and sanity check frame counts . Hardcode for now.
+    ffmpeg -loglevel warning -i "$1"/../transcoder/test.ts -c:a copy -c:v copy -f hls test.m3u8
+    ffprobe -loglevel warning -select_streams v -count_frames -show_streams test0.ts | grep nb_read_frames=120
+    ffprobe -loglevel warning -select_streams v -count_frames -show_streams test1.ts | grep nb_read_frames=120
+    ffprobe -loglevel warning -select_streams v -count_frames -show_streams test2.ts | grep nb_read_frames=120
+    ffprobe -loglevel warning -select_streams v -count_frames -show_streams test3.ts | grep nb_read_frames=120
+  `
+	run(cmd)
+
+	err := ffmpeg.InitFFmpegWithDetectorProfile(&ffmpeg.DSceneAdultSoccer, "0")
+	if err != nil {
+		t.Error(err)
+	}
+	defer ffmpeg.ReleaseFFmpegDetectorProfile()
+	// Test encoding with only seg0 and seg2 under detection
+	tc := NewTranscoder()
+	prof := P144p30fps16x9
+	for i := 0; i < 4; i++ {
+		in := &TranscodeOptionsIn{
+			Fname: fmt.Sprintf("%s/test%d.ts", dir, i),
+			Accel: accel,
+		}
+		out := []TranscodeOptions{
+			{
+				Oname:   fmt.Sprintf("%s/out%d.ts", dir, i),
+				Profile: prof,
+				Accel:   accel,
+			},
+		}
+		if i%2 == 0 {
+			out = append(out, TranscodeOptions{
+				Detector: &DSceneAdultSoccer,
+				Accel:    accel,
+			})
+		}
+		res, err := tc.Transcode(in, out)
+		if err != nil {
+			t.Error(err)
+		}
+		if i%2 == 0 && (len(res.Encoded) < 2 || res.Encoded[1].DetectData == nil) {
+			t.Error("No detect data returned for detection profile")
+		}
+	}
+	tc.StopTranscoder()
+}
+
+func TestTranscoder_DetectionFreq(t *testing.T) {
+	detectionFreq(t, Software)
+}
+*/
 
 /*
 func noKeyframeSegment(t *testing.T, accel Acceleration) {
--- a/ffmpeg/ffmpeg.go
+++ b/ffmpeg/ffmpeg.go
@@ -32,6 +32,7 @@ var ErrTranscoderPrf = errors.New("TranscoderUnrecognizedProfile")
 var ErrTranscoderGOP = errors.New("TranscoderInvalidGOP")
 var ErrTranscoderDev = errors.New("TranscoderIncompatibleDevices")
 var ErrEmptyData = errors.New("EmptyData")
+var ErrDNNInitialize = errors.New("TranscoderInvalidDNNProfile")
 
 type Acceleration int
 
@@ -61,10 +62,11 @@ type TranscodeOptionsIn struct {
 }
 
 type TranscodeOptions struct {
-	Oname   string
-	Profile VideoProfile
-	Accel   Acceleration
-	Device  string
+	Oname    string
+	Profile  VideoProfile
+	Detector DetectorProfile
+	Accel    Acceleration
+	Device   string
 
 	Muxer        ComponentOptions
 	VideoEncoder ComponentOptions
@@ -72,8 +74,9 @@ type TranscodeOptions struct {
 }
 
 type MediaInfo struct {
-	Frames int
-	Pixels int64
+	Frames     int
+	Pixels     int64
+	DetectData DetectData
 }
 
 type TranscodeResults struct {
@@ -246,6 +249,13 @@ func (t *Transcoder) Transcode(input *TranscodeOptionsIn, ps []TranscodeOptions)
 	}
 	params := make([]C.output_params, len(ps))
 	for i, p := range ps {
+		if p.Detector != nil {
+			// We don't do any encoding for detector profiles
+			// Adding placeholder values to pass checks for these everywhere
+			p.Oname = "/dev/null"
+			p.Profile = P144p30fps16x9
+			p.Muxer = ComponentOptions{Name: "mpegts"}
+		}
 		oname := C.CString(p.Oname)
 		defer C.free(unsafe.Pointer(oname))
 
@@ -294,6 +304,19 @@ func (t *Transcoder) Transcode(input *TranscodeOptionsIn, ps []TranscodeOptions)
 		if param.Framerate > 0 {
 			filters += fmt.Sprintf(",fps=%d/%d", param.Framerate, param.FramerateDen)
 			fps = C.AVRational{num: C.int(param.Framerate), den: C.int(param.FramerateDen)}
+		}
+		// if has a detector profile, ignore all video options
+		if p.Detector != nil {
+			deviceid := "0"
+			if input.Accel != Software && len(input.Device) > 0 {
+				deviceid = input.Device
+			}
+			switch p.Detector.Type() {
+			case SceneClassification:
+				detectorProfile := p.Detector.(*SceneClassificationProfile)
+				filters = fmt.Sprintf("lvpdnn=filter_type=lvpclassify:device=%s:model=%s:input=%s:output=%s:sample=%d",
+					deviceid, detectorProfile.ModelPath, detectorProfile.Input, detectorProfile.Output, detectorProfile.SampleRate)
+			}
 		}
 		var muxOpts C.component_opts
 		var muxName string
@@ -419,6 +442,18 @@ func (t *Transcoder) Transcode(input *TranscodeOptionsIn, ps []TranscodeOptions)
 			Frames: int(r.frames),
 			Pixels: int64(r.pixels),
 		}
+		// add detect result
+		if ps[i].Detector != nil {
+			switch ps[i].Detector.Type() {
+			case SceneClassification:
+				detector := ps[i].Detector.(*SceneClassificationProfile)
+				res := make(SceneClassificationData)
+				for j, class := range detector.Classes {
+					res[class.ID] = float64(r.probs[j])
+				}
+				tr[i].DetectData = res
+			}
+		}
 	}
 	dec := MediaInfo{
 		Frames: int(decoded.frames),
@@ -471,4 +506,34 @@ func InitFFmpegWithLogLevel(level LogLevel) {
 
 func InitFFmpeg() {
 	InitFFmpegWithLogLevel(FFLogWarning)
+}
+
+func ReleaseFFmpegDetectorProfile() {
+	C.lpms_dnnrelease()
+}
+func InitFFmpegWithDetectorProfile(detector DetectorProfile, deviceids string) error {
+
+	switch detector.Type() {
+	case SceneClassification:
+		detectorProfile := detector.(*SceneClassificationProfile)
+		dnnOpt := &C.lvpdnn_opts{
+			modelpath:  C.CString(detectorProfile.ModelPath),
+			inputname:  C.CString(detectorProfile.Input),
+			outputname: C.CString(detectorProfile.Output),
+			deviceids:  C.CString(deviceids),
+		}
+		defer C.free(unsafe.Pointer(dnnOpt.modelpath))
+		defer C.free(unsafe.Pointer(dnnOpt.inputname))
+		defer C.free(unsafe.Pointer(dnnOpt.outputname))
+		defer C.free(unsafe.Pointer(dnnOpt.deviceids))
+
+		ret := int(C.lpms_dnninit(dnnOpt))
+		if 0 != ret {
+			glog.Error("lpms_dnninit Return : ", ErrorMap[ret])
+			return ErrDNNInitialize
+		}
+	}
+
+	InitFFmpegWithLogLevel(FFLogWarning)
+	return nil
 }
--- a/ffmpeg/filter.h
+++ b/ffmpeg/filter.h
@@ -60,6 +60,8 @@ struct output_ctx {
 
   int64_t gop_time, gop_pts_len, next_kf_pts; // for gop reset
 
+  int is_dnn_profile; //if not dnn profile: 0
+
   output_results  *res; // data to return for this output
 
 };
--- a/ffmpeg/transcoder.c
+++ b/ffmpeg/transcoder.c
@@ -6,6 +6,8 @@
 
 #include <libavcodec/avcodec.h>
 #include <libavformat/avformat.h>
+#include <libavfilter/avfilter.h>
+#include <stdbool.h>
 
 // Not great to appropriate internal API like this...
 const int lpms_ERR_INPUT_PIXFMT = FFERRTAG('I','N','P','X');
@@ -81,6 +83,19 @@ struct transcode_thread {
 void lpms_init(enum LPMSLogLevel max_level)
 {
   av_log_set_level(max_level);
+}
+
+int lpms_dnninit(lvpdnn_opts *dnn_opts) {
+
+    int res = avfilter_register_lvpdnn(dnn_opts->modelpath,dnn_opts->inputname,dnn_opts->outputname,dnn_opts->deviceids);
+    if(res != 0) {
+      LPMS_WARN("Could not initialize dnn module!");
+    }
+    return res;
+}
+
+void lpms_dnnrelease() {
+  avfilter_remove_lvpdnn();
 }
 
 //
@@ -159,6 +174,7 @@ int transcode(struct transcode_thread *h,
       octx->audio = &params[i].audio;
       octx->video = &params[i].video;
       octx->vfilters = params[i].vfilters;
+      octx->is_dnn_profile = (strncmp(octx->vfilters,LVPDNN_FILTER_NAME, strlen(LVPDNN_FILTER_NAME)) == 0 );      
       if (params[i].bitrate) octx->bitrate = params[i].bitrate;
       if (params[i].fps.den) octx->fps = params[i].fps;
       if (params[i].gop_time) octx->gop_time = params[i].gop_time;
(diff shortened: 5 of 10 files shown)


yondonfu:
@oscar-davids I know you mentioned that the [ffmpeg fork with the detection filter](https://github.com/livepeer/FFmpeg/pull/1) should be built with the `-enable-libtensorflow` flag, but are there any other flags that need to be included when running `./configure` before make + make install? I think it would be helpful for everyone if the exact build steps are posted just so everyone is on the same page.

oscar-davids:
I have tested with the following configure and build command.

 ./configure --enable-libx264 --enable-gnutls --enable-gpl --enable-static --enable-libtensorflow --enable-cuda --enable-cuda-llvm --enable-cuvid --enable-nvenc --enable-decoder=h264_cuvid --enable-filter=scale_cuda --enable-encoder=h264_nvenc

 make
 make install

ffmpeg/filter.h
```diff
@@ -59,7 +59,9 @@ struct output_ctx {
   int64_t drop_ts;     // preroll audio ts to drop
 
   int64_t gop_time, gop_pts_len, next_kf_pts; // for gop reset
-
+#ifdef USE_LVPDNN_
+  int dummy;
```
jailuthra, line 63: Can be changed to a more descriptive variable name like `is_dnn_profile` or similar

ffmpeg/ffmpeg.go
```diff
@@ -68,6 +68,9 @@ type TranscodeOptions struct {
 type MediaInfo struct {
 	Frames int
 	Pixels int64
+	//for dnn result
+	Isdnn     bool
+	Dnnresult string
```
jailuthra, line 73: Can be changed to `IsDNN` and `DNNResult` to be consistent with naming of other variables.

ffmpeg/videoprofile.go
```diff
@@ -36,6 +36,25 @@ const (
...
+)
+
+type DetectorProfile struct {
+	Dnntype    int  //classification or object detect
+	SampleRate uint //execute dnn filter per samplerate
+	ModelPath  string
+	Threshold  float32
+	Input      string
+	Output     string
+	Gpuid      int
```
jailuthra, line 53: `Dnntype` and `Gpuid` -> `DNNType` and `GPUId`

ffmpeg/videoprofile.go
```diff
@@ -73,6 +93,12 @@ var (
 	P240p30fps4x3  = VideoProfile{Name: "P240p30fps4x3", Bitrate: "600k", Framerate: 30, AspectRatio: "4:3", Resolution: "320x240"}
 	P144p30fps16x9 = VideoProfile{Name: "P144p30fps16x9", Bitrate: "400k", Framerate: 30, AspectRatio: "16:9", Resolution: "256x144"}
 	P144p25fps16x9 = VideoProfile{Name: "P144p25fps16x9", Bitrate: "400k", Framerate: 25, AspectRatio: "16:9", Resolution: "256x144"}
+	PDnnDetector   = VideoProfile{Name: "PDnnDetector", Bitrate: "400k", Framerate: 20, AspectRatio: "1:1", Resolution: "224x224",
+		Detector: DetectorProfile{SampleRate: 30, ModelPath: "tasmodel.pb", Threshold: 0.8, Input: "input_1", Output: "reshape_3/Reshape",
+			ClassIds: []int{1}, ClassNames: []string{"adult", "soccer"}}}
+	PDnnVioFilter = VideoProfile{Name: "PDnnVioFilter", Bitrate: "400k", Framerate: 20, AspectRatio: "1:1", Resolution: "224x224",
+		Detector: DetectorProfile{SampleRate: 30, ModelPath: "tviomodel.pb", Threshold: 0.8, Input: "input_1", Output: "reshape_3/Reshape",
+			ClassIds: []int{0}, ClassNames: []string{"violence"}}}
```
jailuthra, line 101: Given that we neither use the output frames or perform any FPS/Scaling filtering when configured with these DNN profiles, does it make sense to remove `Bitrate`, `Framerate`, `Resolution` and other redundant options from here to avoid confusion for the reader?

jailuthra (changes requested):
LGTM after minor fixes.

Given that the output frames after filtering are not useful to us, would it make sense to skip encoding when the DNN filter is used? It could be done via making the AVFilter a sink or adding a check in the C code. This is non-urgent though, and can be done later too. Let me know your thoughts

ffmpeg/ffmpeg.go
yondonfu: What is the format of `Dnnresult`? Would be good to document.

If it is an empty string can we assume that the lvpdnn filter was not used? If so, could we remove `Isdnn` and just check if `len(Dnnresult) == 0` to determine if a DNN result was returned or not?

ffmpeg/ffmpeg.go
```diff
@@ -372,6 +382,17 @@ func (t *Transcoder) Transcode(input *TranscodeOptionsIn, ps []TranscodeOptions)
 			Frames: int(r.frames),
 			Pixels: int64(r.pixels),
 		}
+		//add dnn result
+		dnnresult := ""
```
yondonfu, line 386: We can declare this in the following if block since we don't need the var unless we're in the if block.

ffmpeg/videoprofile.go
yondonfu: Do we need `Gpuid`? Any calls to the `Transcode()` API should already be associated with a specific GPU so we could just run the DNN filter on the GPU that has already been selected.

ffmpeg/videoprofile.go
```diff
@@ -36,6 +36,25 @@ const (
 	GOPInvalid = -2
 )
 
+// for sceneclassification
+const (
```
yondonfu, line 40:
```suggestion
type DNNFilterType int

const (
    DNNFilterType DNNClassify = iota
    ...
```

Since there is a concept of "filter type" in https://github.com/livepeer/FFmpeg/pull/1 we could use a custom enum type to represent that.

ffmpeg/videoprofile.go
```diff
@@ -36,6 +36,25 @@ const (
...
+	DnnClassify = iota
+	DnnYolo
```
yondonfu, line 42: livepeer/FFmpeg#1 only mentions the lvpclassify and lvpodetect filter types. Do we support these additional types as well?

ffmpeg/ffmpeg.go
```diff
@@ -251,6 +254,13 @@ func (t *Transcoder) Transcode(input *TranscodeOptionsIn, ps []TranscodeOptions)
 			filters += fmt.Sprintf(",fps=%d/%d", param.Framerate, param.FramerateDen)
 			fps = C.AVRational{num: C.int(param.Framerate), den: C.int(param.FramerateDen)}
 		}
+		//if has dnn filter, ignore all video options
+		if len(p.Profile.Detector.ModelPath) > 0 {
+			dnnProfile := p.Profile.Detector
+			//filters = "lvpdnn=model=tafmodel.pb:input=input_1:output=reshape_3/Reshape:sample=20:threshold=0.5"
+			filters = fmt.Sprintf("lvpdnn=model=%s:input=%s:output=%s:sample=%d:threshold=%.2f",
```
yondonfu, line 261: Do we need to specify the filter type? i.e. lvpclassify or lvpodetect?

ffmpeg/videoprofile.go
yondonfu: Separate question - should the values for this enum be strings since the filter type described in the ffmpeg PR are strings i.e. "lvpclassify" and "lvpodetect"

ffmpeg/filter.h
oscar-davids: changed to is_dnn_profile.

ffmpeg/videoprofile.go
oscar-davids: I agree, removed Bitrate, Framerate, Resolution, AspectRatio.

ffmpeg/logging.h
```diff
@@ -1,6 +1,13 @@
 #ifndef _LPMS_LOGGING_H_
 #define _LPMS_LOGGING_H_
 
+//MACRO FOR DEVELOP
+#define USE_LVPDNN_
+#ifdef USE_LVPDNN_
```
yondonfu, line 6: Is the main purpose of this macro to hardcode the variables below during development? And this macro wouldn't be used in production?

ffmpeg/transcoder.c
```diff
@@ -271,9 +274,21 @@ int transcode(struct transcode_thread *h,
...
-    ret = flush_outputs(ictx, &outputs[i]);
+  for (i = 0; i < nb_outputs; i++) {  
+#ifdef USE_LVPDNN_  
+    if(outputs[i].is_dnn_profile == 0 ){
+      ret = flush_outputs(ictx, &outputs[i]);
+      if (ret < 0) LPMS_ERR(transcode_cleanup, "Unable to fully flush outputs")
+    } 
+    else if(outputs[i].is_dnn_profile && outputs[i].res->frames > 0){      
+       for (int j = 0; j < MAX_CLASSIFY_SIZE; j++){
+         outputs[i].res->probs[j] =  outputs[i].res->probs[j] / outputs[i].res->frames;
```
yondonfu, line 285: Just to make sure I understand this correctly:

- There will be `j` probability values where j = the # class IDs for the lvpdnn filter
- The jth probability is the probability score for the jth class ID
- We divide the jth probability by the total frames sampled

Is that right?

yondonfu:
I think it will be helpful to make sure that the data structure definitions in this PR match the expectations described in https://github.com/livepeer/go-livepeer/issues/1881. Open to feedback on the data structure definitions in that spec, but assuming that we use those definitions, I suggest the following updates to this PR.

Create a `SceneClassificationProfile` struct that implements the `DetectorProfile` interface which could be used to represent other detector profile implementations later on.

```golang
type DetectorType int

const (
	DetectorType SceneClassification = iota
         // Example for future:
	// ObjectDetection
)

type DetectorProfile interface {
	Type() DetectorType
}

type DetectorClass struct {
    ID int
    string name
}

type SceneClassificationProfile struct {
	SampleRate uint
         ModelPath string
         Input string
         Output string
         Classes []DetectorClass
}

func (p *SceneClassificationProfile) Type() DetectorType {
	return SceneClassification
}
```

Then, instead of nesting the `DetectorProfile` in the `VideoProfile` struct we can add it in the `TranscodeOptions` struct to create more separation between video profiles that describe the configuration for an output rendition and detector profiles that describe the configuration for detection to be run on inputs *without* producing an output rendition.

```golang
type TranscodeOptions struct {
    ...
    Profile VideoProfile
    DetectorProfile DetectorProfile
}
```

The filter creation in the `Transcode()` in `ffmpeg.go` can be updated to cast the `DetectorProfile`, if it is present, into a `SceneClassificationProfile` and in the future other detector profile implementations.

```golang
if p.DetectorProfile != nil {
    switch p.DetectorProfile.Type() {
        case SceneClassificationProfile:
            // Configure scene classification filter
            filters = ...
        default:
            glog.Warningf("Unknown detector profile - skipping")
    }
}
```

For processing detection results, we can use a similar approach as the one used for detector profiles. 

A `DetectData` interface can be used with the `SceneClassificationData` struct implementing the interface.

```golang
type DetectData interface {
    Type() DetectorType
}

type SceneClassificationData map[int]float64

func (scd SceneClassificationData) Type() DetectorType {
    return SceneClassification
}
```

Then, the `MediaInfo` struct can just include the `DetectData`.

```golang
type MediaInfo struct {
    Frames int
    Pixels int64
    DetectData DetectData 
}
```

When the `Transcode()` call creates the detect data, if the detector profile is a scene classification profile, it can create a `SceneClassificationData` that stores a map of class ID -> probability which is added to the `MediaInfo` struct.

```golang
if ps[i].DetectorProfile != nil {
    switch p.DetectorProfile.Type() {
        case SceneClassificationProfile:
            scd := make(map[int]float64)
            scp := p.DetectorProfile.(SceneClassificationProfile)
            for c, j := range scp.Classes {
                scd[c.ID] = r.probs[j]
            }

            tr[i].DetectData = scd
        default:
            glog.Warningf("Unknown detector profile - skipping")
    }
}
```

What do you think?

ffmpeg/transcoder.c
yondonfu: Bump

ffmpeg/logging.h
yondonfu: Bump

jailuthra:
**Detection Freq**

We only want to run detection on `1/Freq` segments of a stream. From T node's point of view, this would look something like -
1. Initialize ffmpeg.Transcoder via `NewTranscoder`
2. Send first segment's `Transcode` request with the output profile that contains `TranscodeOptions.DetectorProfile`
3. Send subsequent `Freq-1` segments without the detection-specific output profile present.
4. Send the `Freq`th segment with the detection-specific output profile present.
5. Back to 3.

With the current state of this PR, turning off detection for some segments mid-stream like described above should theoretically work with minimal changes.

Removing the detection-specific output profile mid-stream might cause an error in `lpms_transcode` which expects number of outputs to remain same https://github.com/livepeer/lpms/blob/f7f2491160259ab365f64c4912da8486925d2854/ffmpeg/transcoder.c#L377-L379

It would be a good idea to start with writing a unit test for this workflow, as other places in the pipeline might also break or throw errors.

Also note, we initialize filtergraph for all the output renditions in `open_output` method:
https://github.com/livepeer/lpms/blob/f7f2491160259ab365f64c4912da8486925d2854/ffmpeg/encoder.c#L204
But this runs only at the start of the session for GPU transcoding:
https://github.com/livepeer/lpms/blob/f7f2491160259ab365f64c4912da8486925d2854/ffmpeg/transcoder.c#L174-L175
This is done as re-initializing filtergraph/encoder on GPU between segments is quite expensive.
Not sure if this will break or not, but let me know if you face any issues testing it out!

**Blocking Changes**

I've started implementing https://github.com/livepeer/go-livepeer/issues/1884 where we're parsing the detector config from the authWebhook's json response into the `DetectorProfile` interface as mentioned by @yondonfu's above https://github.com/livepeer/lpms/pull/228#issuecomment-844257034.

Currently using the latest commit in this PR https://github.com/livepeer/lpms/pull/228/commits/f6c2f98ca71d7b546781c2e8551c80c8343ad280, but noting a soft-blocker on these changes specifically -
1. Same `DetectorProfile` interface for different `DetectorTypes`
2. Using the newer `[]Classes` field directly instead of parsing into `ClassIds` and `ClassNames` slices separately

(and similarly `DetectData` changes are soft-blocker for https://github.com/livepeer/go-livepeer/issues/1885)

cmd/transcoding/transcoding.go
```diff
@@ -77,6 +79,10 @@ func main() {
 	}
 	fmt.Printf("profile=input frames=%v pixels=%v\n", res.Decoded.Frames, res.Decoded.Pixels)
 	for i, r := range res.Encoded {
-		fmt.Printf("profile=%v frames=%v pixels=%v\n", profiles[i].Name, r.Frames, r.Pixels)
+		if r.DetectData != nil {
+			fmt.Printf("profile=%v frames=%v pixels=%v dnnfilter result =%v\n", profiles[i].Name, r.Frames, r.Pixels, r.DetectData)
```
yondonfu, line 83:
```suggestion
			fmt.Printf("profile=%v frames=%v pixels=%v detectdata=%v\n", profiles[i].Name, r.Frames, r.Pixels, r.DetectData)
```

ffmpeg/api_test.go
```diff
@@ -1244,6 +1244,54 @@ func TestTranscoder_AudioOnly(t *testing.T) {
...
+			{
+				Oname:   fmt.Sprintf("%s/out%d.ts", dir, i),
+				Profile: prof,
+				Accel:   accel,
+			},
+		}
+		if i%2 == 0 {
+			out = append(out, TranscodeOptions{
+				Oname:    fmt.Sprintf("%s/out-dnn%d.ts", dir, i),
+				Profile:  prof,
```
yondonfu, line 1278: Could `Profile` be omitted here since populating `Detector` which means we're not creating an output?

ffmpeg/api_test.go
```diff
@@ -1244,6 +1244,54 @@ func TestTranscoder_AudioOnly(t *testing.T) {
...
+		out := []TranscodeOptions{
+			{
+				Oname:   fmt.Sprintf("%s/out%d.ts", dir, i),
+				Profile: prof,
+				Accel:   accel,
+			},
+		}
+		if i%2 == 0 {
+			out = append(out, TranscodeOptions{
+				Oname:    fmt.Sprintf("%s/out-dnn%d.ts", dir, i),
```
yondonfu, line 1277: Is `Oname` still required if `Detector` is used?

ffmpeg/api_test.go
```diff
@@ -1244,6 +1244,54 @@ func TestTranscoder_AudioOnly(t *testing.T) {
 	audioOnlySegment(t, Software)
 }
 
+func detectionFreq(t *testing.T, accel Acceleration) {
```
yondonfu, line 1247: Wouldn't this test always fail CI right now since there is no model available? I guess we can make sure the test can be run locally and just comment it out.

ffmpeg/ffmpeg.go
```diff
@@ -251,6 +254,15 @@ func (t *Transcoder) Transcode(input *TranscodeOptionsIn, ps []TranscodeOptions)
 			filters += fmt.Sprintf(",fps=%d/%d", param.Framerate, param.FramerateDen)
 			fps = C.AVRational{num: C.int(param.Framerate), den: C.int(param.FramerateDen)}
 		}
+		//if has dnn filter, ignore all video options
```
yondonfu, line 257:
```suggestion
		// if has a detector profile, ignore all video options
```

ffmpeg/ffmpeg.go
```diff
@@ -251,6 +254,15 @@ func (t *Transcoder) Transcode(input *TranscodeOptionsIn, ps []TranscodeOptions)
...
+		if p.Detector != nil {
+			switch p.Detector.Type() {
+			case SceneClassification:
+				dnnProfile := p.Detector.(*SceneClassificationProfile)
```
yondonfu, line 261:
```suggestion
				detectorProfile := p.Detector.(*SceneClassificationProfile)
```

ffmpeg/ffmpeg.go
```diff
@@ -372,6 +384,18 @@ func (t *Transcoder) Transcode(input *TranscodeOptionsIn, ps []TranscodeOptions)
 			Frames: int(r.frames),
 			Pixels: int64(r.pixels),
 		}
+		//add dnn result
```
yondonfu, line 387:
```suggestion
		// add detect result
```

ffmpeg/transcoder.c
```diff
@@ -159,6 +160,9 @@ int transcode(struct transcode_thread *h,
       octx->audio = &params[i].audio;
       octx->video = &params[i].video;
       octx->vfilters = params[i].vfilters;
+#ifdef USE_LVPDNN_
+      octx->is_dnn_profile = (strncmp(octx->vfilters,LVPDNN_FILTER_NAME, strlen(LVPDNN_FILTER_NAME)) == 0 );
```
yondonfu, line 164:
```suggestion
      octx->is_detector_profile = (strncmp(octx->vfilters,LVPDNN_FILTER_NAME, strlen(LVPDNN_FILTER_NAME)) == 0 );      
```

yondonfu (changes requested):

yondonfu:
@oscar-davids @jailuthra I want us to get on the same page regarding a few things for this PR:

## Known Issues

**Address Later**

- **Copying data between GPU & CPU memory**: At the moment, the models are trained on static image data. So, running inference with the models requires static image data as input. As a result, the lvpdnn filter currently copies decoded video frames into CPU memory and converts them into a format that is compatible with the trained models. We know that this is sub-optimal because a) it adds additional overhead likely slowing down the pipeline and b) it imposes an additional memory burden on the CPU which can be a problem in environments without access to high performance CPUs. After a discussion with @oscar-davids, I think this is ok *for now* and can be addressed later, but if testing results (see the next section) indicate that we need performance improvements then we'll have to re-visit this sooner.

**Address Now**

- **GPU selection**: At the moment, the lvpdnn filter only uses GPU 0 because the `gpu_options.visible_device_list` configuration for the TF session is hardcoded with GPU 0. Under different circumstances, this would mean that if a segment was transcoded on any other GPU, then scene classification could not be run for that segment. However, since the lvpdnn filter currently copies data between GPU & CPU memory, currently a segment transcoded on another GPU could have the necessary data copied to CPU and then uploaded to GPU 0 for scene classification. Nevertheless, we should aim to run scene classification on the same GPU that is being used for transcoding instead of only using a single GPU to make sure we're utilizing all GPUs available on a machine [1]. I've written up a [suggestion](https://github.com/livepeer/FFmpeg/pull/1#discussion_r640910389) for how to address this issue.
- **Model loading**: At the moment, a model is loaded into GPU VRAM when the lvpdnn filter is initialized which would happen for the first segment in a stream. Loading a model can take a long as 1 second. The first segment in a stream is already slower to transcode to the overhead from initializing encoder sessions so we really want to avoid adding any additional overhead especially since the overhead for loading a model is pretty high. Let's think about ways to avoid incurring the model loading overhead for the first segment of a stream. My first thought was to pre-load the model (i.e. in go-livepeer this could be on startup). But, this approach might not be compatible with the suggestion for the previous issue of configuring the specific GPU to use when loading a model.

[1] Plus, presumably we will eventually remove data copying between GPU and CPU memory. At that point, we definitely would need scene classification to run on the same GPU as transcoding.

**Testing**

I think we should understand the following:

1. For segments that require scene classification, how long does scene classification take for various values of the `SampleRate` (per frame sampling) parameter defined in the `SceneClassificationProfile`?
2. For segments that require scene classification, what is the real-time ratio for a segments for various values of the `SampleRate` (per frame sampling) parameter defined in the `SceneClassificationProfile`?
3. For segments that require scene classification, what is the accuracy for each segment?

ffmpeg/api_test.go
jailuthra: Good idea - note to future-self, move this and the keyframe test in a separate file with `//+build ignored`

ffmpeg/api_test.go
jailuthra: Fixed in https://github.com/livepeer/lpms/pull/228/commits/8eb36316acb27b953a5920e0a49c0fe40747651d

ffmpeg/videoprofile.go
```diff
@@ -36,6 +36,25 @@ const (
...
+	DnnMasking
+)
+
+type DetectorProfile struct {
+	Dnntype    int  //classification or object detect
+	SampleRate uint //execute dnn filter per samplerate
+	ModelPath  string
+	Threshold  float32
```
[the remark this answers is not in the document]
jailuthra: Just noticed I did this in https://github.com/livepeer/lpms/pull/228/commits/b8ec5195511449a56cd0436e07c2917f0f281973 - resolving comment

jailuthra:
> 1. For segments that require scene classification, how long does scene classification take for various values of the `SampleRate` (per frame sampling) parameter defined in the `SceneClassificationProfile`?
> 2. For segments that require scene classification, what is the real-time ratio for a segments for various values of the `SampleRate` (per frame sampling) parameter defined in the `SceneClassificationProfile`?
> 3. For segments that require scene classification, what is the accuracy for each segment?

Made some updates to the `livepeer_bench` tool https://github.com/livepeer/go-livepeer/commit/cd7d4d4d5b1b1ba75f9762c09184673355519415 for testing these metrics

This [gist](https://gist.github.com/jailuthra/5392aa27019c66970ffdd6b4a1dee6d6) contains a bash script and raw benchmarking outputs for different configurations of content-detection.

Eye-balling some key metrics:

**First segment transcode time:**
```
No Detection                         - 300ms
Detection with 1 concurrent sessions - 4500ms
Detection with 2 concurrent sessions - 4500ms (for sess #1), 2500ms (for sess #2)
```
With frame sampling rate set at 1 (i.e. detection runs at every frame) an additional ~1s to each of these ^

**Approximate segment Transcode Duration and Real-Time ratio:**
```
No Detection                   - 100ms  - RT 0.05
Detection on every (1st) frame - 1000ms - RT 0.49
Detection on every 5th frame   - 240ms  - RT 0.18
Detection on every 10th frame  - 180ms  - RT 0.14
Detection on every 30th frame  - 120ms  - RT 0.11
```

**Accuracy** of classifying a segment as soccer was pretty good regardless of the frame-sampling-rate.
Even when sampling every 30th frame, soccer classification probability was > 0.5 for all relevant segments.

ffmpeg/ffmpeg.go
```diff
@@ -419,3 +455,33 @@ func InitFFmpegWithLogLevel(level LogLevel) {
 func InitFFmpeg() {
 	InitFFmpegWithLogLevel(FFLogWarning)
 }
+
+func ReleaseFFmpeg() {
+	C.lpms_dnnrelease()
+}
+func InitFFmpegWithDetectProfile(detector DetectorProfile, deviceids string) error {
```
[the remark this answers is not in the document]
jailuthra: Nit: Rename to `InitFFmpegWithDetectorProfile`

jailuthra (commented):
Minor suggestions, other than that LGTM.

ffmpeg/transcoder.c
```diff
@@ -6,6 +6,7 @@
 
 #include <libavcodec/avcodec.h>
 #include <libavformat/avformat.h>
+#include <stdbool.h>
```
yondonfu, line 9: I think we might need to include avfilter.h here - see [here](https://github.com/livepeer/lpms/commit/7f3f7c2881a63e56be53e116dbc030d4e3d3d9d9).

Without the include statements, I ran into compilation errors about missing references for `avfilter_register_lvpdnn` and `avfilter_remove_lvpdnn`.

yondonfu (changes requested):
I think CI tests will fail unless `install_ffmpeg.sh` is updated to point to the ffmpeg commit that contains the lvpdnn filter - the update would look like [this](https://github.com/livepeer/lpms/commit/9ceb55e4d76f32432b9fc309348b7f77572bd3f2). We missed this up until now because this PR was opened via a fork.

After https://github.com/livepeer/FFmpeg/pull/1 is merged, let's make sure to update `install_ffmpeg.sh` to point to the latest ffmpeg commit on the `livepeer` branch.

I also switched the target branch for this PR to the master branch.

jailuthra (approved):
LGTM for merge, after updating the `install_ffmpeg.sh` script & including `avfilter.h` as mentioned by Yondon.

We can fix some of my other comments separately, and consider them as a short-term tech debt.

cmd/scenedetection/scenedetection.go
```diff
@@ -0,0 +1,108 @@
...
+				Oname:   fmt.Sprintf("out_%s_%d_out.mkv", lbl, i),
+				Profile: profs[i],
+				Accel:   accel,
+			}
+			opts = append(opts, o)
+		}
+		//add detection profile
+		o := ffmpeg.TranscodeOptions{
+			Oname:    "out_null.mkv",
+			Profile:  ffmpeg.P144p30fps16x9,
```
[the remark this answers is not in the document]
oscar-davids: I have some tests in windows env,  there  p.Oname = "/dev/null" was not working

cmd/scenedetection/scenedetection.go
```diff
@@ -0,0 +1,108 @@
...
+	valids := make([]string, len(ffmpeg.VideoProfileLookup))
+	for p, _ := range ffmpeg.VideoProfileLookup {
+		valids = append(valids, p)
+	}
+	return valids
+}
+
+func main() {
+	if len(os.Args) <= 4 {
+		//0,1 input.mp4 P720p25fps16x9,P720p30fps4x3 nv 0
```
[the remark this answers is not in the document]
oscar-davids: Good Idea, I thought that this example is effective for pre-loading test. currently, I don't want to mix with software decoding and content detection attribute.

ffmpeg/transcoder.c
oscar-davids: Updated install_ffmpeg.sh, added avfilter.h head file

ffmpeg/api_test.go
```diff
@@ -1244,6 +1244,57 @@ func TestTranscoder_AudioOnly(t *testing.T) {
 	audioOnlySegment(t, Software)
 }
 
+/*
+func detectionFreq(t *testing.T, accel Acceleration) {
+	run, dir := setupTest(t)
```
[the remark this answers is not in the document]
oscar-davids: Added  InitFFmpegWithDetectorProfile function in test case.

install_ffmpeg.sh
```diff
@@ -24,10 +24,15 @@ if [ ! -e "$HOME/x264/x264" ]; then
 fi
 
 if [ ! -e "$HOME/ffmpeg/libavcodec/libavcodec.a" ]; then
-  git clone https://github.com/livepeer/FFmpeg.git "$HOME/ffmpeg" || echo "FFmpeg dir already exists"
+  LIBTENSORFLOW_VERSION=2.3.0 \
```
yondonfu, line 27: cc @jailuthra for thoughts on whether this should be always installed or whether it should be conditional similar to the approach that is being worked on for https://github.com/livepeer/go-livepeer/issues/1919.

jailuthra: Yeah makes sense to keep it conditional on `BUILD_TAGS="experimental"`.

Also looks like this is not really needed to get the CI build+test scripts running (https://github.com/livepeer/lpms/tree/yf/adddnn) - So I'm fine if we comment it out altogether for now, and handle this later when we enable the unit tests and fix the build process.

cmd/scenedetection/scenedetection.go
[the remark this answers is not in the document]
jailuthra: > I have some tests in windows env, there p.Oname = "/dev/null" was not working
Hmm that's a good catch - we will need to handle that in a more platform independent way before we release this for public network use. Fine for now :+1:

install_ffmpeg.sh
jailuthra: @oscar-davids What do you think about commenting this out for now and moving ahead for the merge?

oscar-davids: That makes sense. So we want to comment all related libtensorflow? Like this?

//&& curl -LO https://storage.googleapis.com/tensorflow/libtensorflow/libtensorflow-cpu-linux- ...
//--enable-libtensorflow

jailuthra: Yep that sounds right. Everything from L27-L30

yondonfu:
After https://github.com/livepeer/FFmpeg/pull/1 is merged let's update the ffmpeg commit hash in install_ffmpeg.sh. Then, this PR should be ready for a merge.

jailuthra:
https://github.com/livepeer/FFmpeg/pull/1 is now merged with this https://github.com/livepeer/FFmpeg/commit/b344789a31a2d806be702239d7614f7a9f51f941 as the final commit that we should use in `install_ffmpeg.sh`

cmd/scenedetection/scenedetection.go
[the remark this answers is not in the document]
jailuthra: Noting for future: `os.DevNull` is Go's platform-independent null device https://golang.org/pkg/os/#pkg-constants

jailuthra:
Once the squashing is complete, let's cherry pick this commit within this PR too: https://github.com/livepeer/lpms/commit/b196bdaaea2a670816cfdbde4f8e35d6aeb51b3a

`git cherry-pick b196bdaaea2a670816cfdbde4f8e35d6aeb51b3a`

These changes (adding a lookup table for models) were needed by the now merged PR https://github.com/livepeer/go-livepeer/pull/1922 for the new API.

ffmpeg/api_test.go
```diff
@@ -1244,6 +1244,8 @@ func TestTranscoder_AudioOnly(t *testing.T) {
 	audioOnlySegment(t, Software)
 }
 
+<<<<<<< HEAD
+<<<<<<< HEAD
```
jailuthra, line 1248: Remove these stray lines from a merge conflict.

ffmpeg/api_test.go
```diff
@@ -1293,6 +1295,64 @@ func TestTranscoder_OutputFPS(t *testing.T) {
 	outputFPS(t, Software)
 }
 
+=======
+>>>>>>> ffmpeg: Comment out detection related tests
```
jailuthra, line 1299: Remove these stray lines from a merge conflict.

jailuthra (changes requested):

ffmpeg/api_test.go
oscar-davids: Removed merge conflict comment.

ffmpeg/api_test.go
oscar-davids: Removed merge conflict comment

jailuthra (approved):
:ship:

PR #231, opened 2021-05-03
README: --tags instead of -tag for Nvidia tests

yondonfu:
`-tag` does not seem to be a valid flag for `go test`. I had to use `--tags=nvidia` in order to run the GPU tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,7 +141,7 @@ this.
 To execute the nvidia tests within the `ffmpeg` directory, run this command:
 
 ```
-go test -tag=nvidia -run Nvidia
+go test --tags=nvidia -run Nvidia
 
 ```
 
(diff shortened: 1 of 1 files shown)


oscar-davids (approved):
I think this flag is really confusing one.
Maybe the flag is dependent on the go version.
As we know we can check how to use this as the following command.

 - go help build

Here they said that "-tags tag" is a comma-separated list, but "--tags" is also worked currently.

jailuthra (approved):
LGTM! thx for fixing this, never noticed it.

yondonfu:
> Here they said that "-tags tag" is a comma-separated list, but "--tags" is also worked currently.

Yeah...I think the `-tag` flag might've worked in a previous Go version as well? Not sure...

Issue #240, opened 2021-06-23
ffmpeg: Better pre-loading API for content detection models


jailuthra:
### Motivation

While implementing content detection we decided to "pre-load" the DNN models at startup after finding out that it is an expensive operation (see https://github.com/livepeer/lpms/pull/228#issuecomment-849914099 for details)

https://github.com/livepeer/lpms/blob/b196bdaaea2a670816cfdbde4f8e35d6aeb51b3a/ffmpeg/ffmpeg.go#L475

This was achieved by triggering a livepeer-specific `libavfilter` API call:
https://github.com/livepeer/lpms/blob/b196bdaaea2a670816cfdbde4f8e35d6aeb51b3a/ffmpeg/transcoder.c#L90
 
This is problematic because:
1. The LPMS-user is also tied to the `InitFFmpeg` API which is a module-level API for initializing things common amongst transcoding sessions (like logging levels), and is ran only once at the node startup in go-livepeer.
2. The FFmpeg codebase is now hard to maintain as we've tweaked the public-level header `avfilter/avfilter.h`directly. It would be preferable to keep our changes contained to our filter module `avfilter/vf_lvpdnn.c`.

#### 1. LPMS API for pre-loading

Expanding on https://github.com/livepeer/lpms/pull/228#discussion_r647464088 :

We can de-couple the model loading from the current `InitFFmpeg` API, and tie it more closely to a "transcode session" that is started up by the LoadBalancer in go-livepeer.

We will still need to map "model A" on suppose "GPU device i" before a stream comes in - but that can be made a runtime decision by the LoadBalancer, rather than a node-level decision taken at the time of node startup.

##### (a) Use binary AVOption

The AVOptions API allows passing and setting custom binary data in an internal context of a codec or filter via https://github.com/livepeer/FFmpeg/blob/livepeer/libavutil/opt.h#L229

There's a `memcpy` going on there so we cannot use it directly for our model, but we can probably pass a double pointer to our model `DNNModel **`  via this API and dereference it in the filter code.

##### (b) Manually cast the priv field 

If we use the `avfilter_graph_create_filter` API instead of `avfilter_graph_parse_ptr` -> we'll get an initialized `AVFilterContext` back.

This should give us access to the `AVFilterContext->priv` field that contains our detection filter's internal context: `LVPDNNContext`.

[...]

PR #246, opened 2021-07-02
ffmpeg: Upgrade to v4.4

jailuthra:
**Motivation**
- See https://github.com/livepeer/internal-project-tracking/issues/141, #247, #248 

**What all has changed?**

*See commit history/comments*


jailuthra:
https://github.com/livepeer/lpms/pull/246/commits/6ccaecaad5f25f7a061ffb5d306aba3827eb64d8 - This commit uses the new scene classification filter introduced in https://github.com/livepeer/FFmpeg/pull/9

It also fixes #240 by introducing a better DNN model pre-loading API

Instead of tying model initializing to a module-wide `InitFFmpeg...` method - We now initialize our scene classification filter while creating a new `Transcoder` session. The DNN filter is then connected to the rest of the filters in the filtergraph when the first segment comes in.

This will require some changes on the go-livepeer load-balancer to do the pre-initialization per transcode session, instead of node-level like done right now.

An added benefit of the new API is that we don't mess with FFmpeg's internal tensorflow model loading APIs, and an `AVFilterContext` is now a self-contained unit of a pre-initialized model for us.

ffmpeg/filter.c
```diff
@@ -204,14 +225,19 @@ int init_audio_filters(struct input_ctx *ictx, struct output_ctx *octx)
...
   if (is_video && inf && inf->hw_frames_ctx && filter->hwframes &&
       inf->hw_frames_ctx->data != filter->hwframes) {
     free_filter(&octx->vf); // XXX really should flush filter first
+    if (octx->dnn_filtergraph) {
+      // swap filtergraph with the pre-initialized DNN filtergraph
+      octx->vf.graph = *octx->dnn_filtergraph;
+    }
     ret = init_video_filters(ictx, octx);
     if (ret < 0) return lpms_ERR_FILTERS;
+    struct filter_ctx *vf = filter;
```
yondonfu, line 240: What does this line do? Doesn't seem like the var is used elsewhere?

ffmpeg/transcoder.h
```diff
@@ -85,10 +86,11 @@ enum LPMSLogLevel {
 void lpms_init(enum LPMSLogLevel max_level);
 int  lpms_transcode(input_params *inp, output_params *params, output_results *results, int nb_outputs, output_results *decoded_results);
 struct transcode_thread* lpms_transcode_new();
+struct transcode_thread* lpms_transcode_new_with_dnn(lvpdnn_opts *dnn_opts);
 void lpms_transcode_stop(struct transcode_thread* handle);
 void lpms_transcode_discontinuity(struct transcode_thread *handle);
 
-int lpms_dnninit(lvpdnn_opts *dnn_opts);
-void lpms_dnnrelease();
+AVFilterGraph * lpms_dnninit(lvpdnn_opts *dnn_opts);
```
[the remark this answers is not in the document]
jailuthra: Good catch will do!

ffmpeg/filter.c
jailuthra: Ah I have no idea when I wrote that, will remove!

ffmpeg/filter.c
```diff
@@ -93,9 +97,26 @@ int init_video_filters(struct input_ctx *ictx, struct output_ctx *octx)
                                     &inputs, &outputs, NULL);
     if (ret < 0) LPMS_ERR(vf_init_cleanup, "Unable to parse video filters desc");
 
+    if (octx->is_dnn_profile && vf->graph == *octx->dnn_filtergraph) {
```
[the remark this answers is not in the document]
jailuthra: This is not a comparison between the filters inside the filtergraph - but a comparison of the `AVFilterGraph *` pointers. 

The purpose of this is a simple validity check, ensuring that we're using the same filtergraph pointer which contains the `livepeer_dnn` filter that we [initialized beforehand](https://github.com/livepeer/lpms/pull/246/commits/6ccaecaad5f25f7a061ffb5d306aba3827eb64d8#diff-7a915a51092546ae69c897ec6a04dd9d49d271ce046de61741f6424ad5218ca8R443). If yes, we'll be able to search for that filter, and connect it at the end of our `hwdownload->format->` filter chain.

ffmpeg/ffmpeg.go
```diff
@@ -359,6 +359,15 @@ func (t *Transcoder) Transcode(input *TranscodeOptionsIn, ps []TranscodeOptions)
 			default:
 				return nil, ErrTranscoderPrf
 			}
+			if p.Profile.Framerate == 0 && p.Accel == Nvidia {
+				// When the decoded video contains non-monotonic increases in PTS (common with OBS)
```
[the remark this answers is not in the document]
jailuthra: That's correct. If FPS passthrough is not enabled we use the `fps` filter, where we manually introduce monotonic timestamps https://github.com/livepeer/lpms/blob/0054a3bd93aa13fbf143946642ece4888a4afcce/ffmpeg/filter.c#L221-L223 (because it doesn't work well without it)

ffmpeg/transcoder.c
```diff
@@ -443,9 +435,62 @@ void lpms_transcode_stop(struct transcode_thread *handle) {
...
+  return NULL;
+}
+
+struct transcode_thread* lpms_transcode_new_with_dnn(lvpdnn_opts *dnn_opts)
+{
+  struct transcode_thread *h = malloc(sizeof (struct transcode_thread));
+  if (!h) return NULL;
+  memset(h, 0, sizeof *h);
+  AVFilterGraph *filtergraph = create_dnn_filtergraph(dnn_opts);
+  if (!filtergraph) return NULL;
```
oscar-davids, line 489: nitpick: needs to free created h handler.

jailuthra: Good catch - this might've lead to a memleak. Will fix!

ffmpeg/encoder.c
```diff
@@ -308,10 +313,9 @@ static int encode(AVCodecContext* encoder, AVFrame *frame, struct output_ctx* oc
     octx->res->pixels += encoder->width * encoder->height;
   }
 
-
   // We don't want to send NULL frames for HW encoding
   // because that closes the encoder: not something we want
-  if (AV_HWDEVICE_TYPE_NONE == octx->hw_type || frame) {
+  if (AV_HWDEVICE_TYPE_NONE == octx->hw_type || AVMEDIA_TYPE_AUDIO == ost->codecpar->codec_type || frame) {
```
oscar-davids, line 318: Here certainly do we need to check AVMEDIA_TYPE_AUDIO , I think that should accept audio NULL frame to flush the encoder.

ffmpeg/ffmpeg.go
```diff
@@ -512,32 +529,42 @@ func InitFFmpeg() {
...
-	return nil
+func deviceidToConfigProto(deviceid string) string {
+	config_proto := &pb.ConfigProto{GpuOptions: &pb.GPUOptions{AllowGrowth: true, VisibleDeviceList: deviceid}}
+	bytes, err := proto.Marshal(config_proto)
+	if err != nil {
+		glog.Errorf("Unable to convert deviceid %v to Tensorflow config protobuf\n", err)
+		return ""
+	}
+	sess_config_option := "sess_config=0x"
+	for i := len(bytes) - 1; i >= 0; i-- {
```
yondonfu, line 566: Is this supposed to parse each byte in `bytes` in reverse order? If so, why do we need to do that?

ffmpeg/ffmpeg.go
```diff
@@ -512,32 +529,42 @@ func InitFFmpeg() {
...
+			}, nil
 		}
 	}
+	return nil, ErrDNNInitialize
+}
 
-	InitFFmpegWithLogLevel(FFLogWarning)
-	return nil
+func deviceidToConfigProto(deviceid string) string {
+	config_proto := &pb.ConfigProto{GpuOptions: &pb.GPUOptions{AllowGrowth: true, VisibleDeviceList: deviceid}}
```
yondonfu, line 559: Nit: Camel case is more conventional.

ffmpeg/filter.c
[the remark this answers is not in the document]
yondonfu: Got it. Thanks for the explanation!

ffmpeg/ffmpeg.go
jailuthra: Yeah this was a pain - I'll add a comment to make this clear.

But somehow FFmpeg tensorflow backend expects the bytes in the reverse order - and their suggested python script even explicitly reverse the bytes in the serialized list: 

https://github.com/livepeer/FFmpeg/blob/c83669bfc0fb683b651b3c354e3f59b736091ef2/libavfilter/dnn/dnn_backend_tf.c#L220

If this wasn't the case we could have used `hex.EncodeToString(bytes)` but oh well :/

jailuthra: Looks like they re-reverse it to construct into a string that `TF_SetConfig` can actually accept. https://github.com/livepeer/FFmpeg/blob/c83669bfc0fb683b651b3c354e3f59b736091ef2/libavfilter/dnn/dnn_backend_tf.c#L251

I'll probably send a patch upstream to make this simpler.

yondonfu:
Changes look good. What's the plan with the ffmpeg dep? Is `jai/ffmpeg-v4.4-working` in ffmpeg going to be merged into `livepeer` before we merge this PR?

jailuthra:
> What's the plan with the ffmpeg dep? Is jai/ffmpeg-v4.4-working in ffmpeg going to be merged into livepeer before we merge this PR?

Yes, I will manually push `jai/ffmpeg-v4.4-working` to `livepeer` and rename the current `livepeer -> livepeer-pre-4.4` to keep an archive around.

yondonfu (approved):
LGTM

oscar-davids (approved):
LGTM!

PR #249, opened 2021-07-08
fix HasZeroVideoFrameBytes for long segments

darkdarkdragon:
ffmpeg doesn't close pipes, and when opening stream
doesn't read all the data - so `io.Copy` was waiting
forever.

### DIFF
--- a/ffmpeg/ffmpeg.go
+++ b/ffmpeg/ffmpeg.go
@@ -103,19 +103,13 @@ func HasZeroVideoFrameBytes(data []byte) (bool, error) {
 	fname := fmt.Sprintf("pipe:%d", or.Fd())
 	cfname := C.CString(fname)
 	defer C.free(unsafe.Pointer(cfname))
-	done := make(chan struct{})
 	go func() {
-		var err2 error
 		br := bytes.NewReader(data)
-		_, err2 = io.Copy(ow, br)
-		if err2 != nil {
-			glog.Errorf("Error sending data to lpms_is_bypass_needed function err=%v", err2)
-		}
+		io.Copy(ow, br)
 		ow.Close()
-		done <- struct{}{}
 	}()
 	bres := int(C.lpms_is_bypass_needed(cfname))
-	<-done
+	ow.Close()
 	return bres == 1, nil
 }
 
--- a/ffmpeg/ffmpeg_test.go
+++ b/ffmpeg/ffmpeg_test.go
@@ -1581,7 +1581,7 @@ func TestTranscoder_ZeroFrame(t *testing.T) {
 	if res != true {
 		t.Errorf("Expecting true, got %v fname=%s", res, fname)
 	}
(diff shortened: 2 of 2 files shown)


ffmpeg/ffmpeg_test.go
```diff
@@ -1591,3 +1591,14 @@ func TestTranscoder_ZeroFrame(t *testing.T) {
...
 	}
 }
+
+func TestTranscoder_ZeroFrameLongBadSegment(t *testing.T) {
+	badSegment := make([]byte, 16*1024*1024)
+	res, err := HasZeroVideoFrameBytes(badSegment)
+	if err != nil {
+		t.Error(err)
+	}
+	if res != false {
```
yondonfu, line 1601:
```suggestion
	if res {
```

ffmpeg/ffmpeg.go
```diff
@@ -103,19 +103,13 @@ func HasZeroVideoFrameBytes(data []byte) (bool, error) {
...
-		if err2 != nil {
-			glog.Errorf("Error sending data to lpms_is_bypass_needed function err=%v", err2)
-		}
+		io.Copy(ow, br)
 		ow.Close()
-		done <- struct{}{}
 	}()
 	bres := int(C.lpms_is_bypass_needed(cfname))
-	<-done
+	ow.Close()
```
yondonfu, line 112: A few questions for my own personal understanding:

- Was the hanging issue caused by `io.Copy` never returning because ffmpeg doesn't read all the data written to the pipe with `ow` so we never reach an EOF while reading from `br`? And this fixes the hanging issue because we no longer wait for the goroutine to complete after `lpms_is_bypass_needed` returns so even if the `io.Copy` in the goroutine hangs we can still return from this function?
- Does the `io.Copy` in the goroutine eventually return?

darkdarkdragon: Yes, `io.Copy` was stuck. I've added `ow.Close` after `lpms_is_bypass_needed` - that makes `io.Copy` to return with `file already closed error`.

yondonfu: Got it. Thanks for the explanation.

And just to double check...its fine to call Close() on `ow` multiple times since we'll call Close() on L109 after `io.Copy` errors out after the first Close() call on L112?

darkdarkdragon: From documentation:
` Close will return an error if it has already been called.`
In my tests it just returns same error `file already closed`
And we need call to `ow.Close` after `io.Copy` because for some segments `lpms_is_bypass_needed` will not exit even though we've sent whole segment already and we need to signal to ffmpeg that there is no more data available.

yondonfu: Answered my own question:

The second Close() will return `close |0: file already closed` which should be fine.

yondonfu (approved):

Issue #251, opened 2021-07-19
Create Go signature comparison function that can pass binary signatures to public ffmpeg API


yondonfu:
- Read signatures
- Load signature data into C structs that can be passed to public ffmpeg API signature comparison function

PR #253, opened 2021-07-26
implementation go signature comparison function

oscar-davids:
Generate Mpeg7 signature binary file while transcoding.
Create Go signature comparison function that can pass binary signatures to public ffmpeg API #251 
Integrate signature filter with unit tests #252

- unit test command

        cd ffmpeg
        go test -run SignData


ffmpeg/signcomp_test.go
```diff
@@ -0,0 +1,71 @@
+package ffmpeg
```
jailuthra, line 1: This file has UNIX permissions set to executable, is that intended?

I also noticed this PR changed `ffmpeg/encoder.c` unix permissions from read/write to read/write/execute `10644 -> 10755`. We should revert that if it was spurious, `chmod -x filename` should do the job I think

ffmpeg/transcoder.c
```diff
@@ -327,6 +329,11 @@ int transcode(struct transcode_thread *h,
     if(outputs[i].is_dnn_profile == 0) {
       ret = flush_outputs(ictx, &outputs[i]);
       if (ret < 0) LPMS_ERR(transcode_cleanup, "Unable to fully flush outputs")
+      //send EOF signal to signature filter
+      if(outputs[i].sfilters != NULL) {
+        av_buffersrc_close(outputs[i].sf.src_ctx, AV_NOPTS_VALUE, AV_BUFFERSRC_FLAG_PUSH);
+        free_filter(&outputs[i].sf);
+      }
```
jailuthra, line 336: We should also ensure the filter is flushed and freed if any error occurs in our transcode loop and we hit `transcode_cleanup`. Maybe moving this whole section inside `transcode_cleanup:` makes more sense.

ffmpeg/sign_test.go
```diff
@@ -0,0 +1,175 @@
...
+		glog.Fatal("Couldn't decode input manifest: ", err)
+	}
+	pl, ok := p.(*m3u8.MediaPlaylist)
+	if !ok {
+		glog.Fatalf("Expecting media playlist in the input %s", in)
+	}
+
+	var wg sync.WaitGroup
+
+	concurrentSessions := 1
```
jailuthra, line 55: Given that we're hardcoding concurrent sessions to 1, we should simplify the loop below, and do away with the use of goroutines and workgroups.

We can also remove any other un-necessary code from the benchmarking module here, like the `time.Sleep`s and extra logging and tablewriter use. Test cases usually don't need to report those details unless they fail.

ffmpeg/signcomp_test.go
```diff
@@ -0,0 +1,71 @@
...
+		for i := 1; i <= 2; i++ {
+			vname := fmt.Sprintf("sign_%s%d.bin", dev, i)
+			fname := path.Join(wd, "..", "data", vname)
+			fnames = append(fnames, fname)
+		}
+	}
+
+	fnames = append(fnames, path.Join(wd, "..", "data", "nodata.bin"))
+
+	res, err := CompareSignatureByPath(fnames[0], fnames[2])
```
jailuthra, line 28: nit: IMO it would be simpler to understand the code if we read the signature binary files into variables that represent what they are, so that our comparisons look like: `CompareSignatureByPath(sw_1_sig, nv_1_sig)` etc.

ffmpeg/ffmpeg.go
```diff
@@ -399,9 +445,11 @@ func (t *Transcoder) Transcode(input *TranscodeOptionsIn, ps []TranscodeOptions)
 			opts: newAVOpts(p.AudioEncoder.Opts),
 		}
 		vfilt := C.CString(filters)
+		sfilt := C.CString(signfilter)
```
jailuthra, line 448: This line and its corresponding free-ing call can be moved inside `if p.CalcSign {` block.

ffmpeg/sign_test.go
```diff
@@ -0,0 +1,175 @@
...
+		{"* Real-Time Duration Ratio *", fmt.Sprintf("%0.4v", transcodeDur/srcDur)},
+	}
+	statsTable.SetAlignment(tablewriter.ALIGN_LEFT)
+	statsTable.SetCenterSeparator("*")
+	statsTable.SetColumnSeparator("|")
+	statsTable.AppendBulk(stats)
+	statsTable.Render()
+}
+
+func TestSign_SWTranscoding(t *testing.T) {
```
jailuthra, line 152: What are these unit-tests testing exactly?

Does it make more sense to have a testcase that generates signatures and compares them, similar to what is done in `signcomp_test.go`?

If the motive here was to not unit test, but have a module that can generate signatures for us - it makes sense to add an option in our transcoding tool inside `cmd/` to optionally generate signatures for us. WDYT?

jailuthra (changes requested):
I had some questions and made some suggestions, but the overall flow of creating a separate filtergraph and passing it scaled frames before encoding LGTM.

ffmpeg/sign_test.go
jailuthra: Also if we decide to keep this as a test-case, we'll need to add a `nvidia` tag in the file, so that our CI pipeline which doesn't have GPUs is able to skip running these testcases.

oscar-davids: Right, I removed benchmarking test code and added all test code(generate and compare) into test_sign.go file.

ffmpeg/filter.c
```diff
@@ -222,6 +222,115 @@ int init_audio_filters(struct input_ctx *ictx, struct output_ctx *octx)
   return ret;
 }
 
+int init_signature_filters(struct output_ctx *octx, AVFrame *inf)
```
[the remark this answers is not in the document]
oscar-davids: I added the common utility function "filtergraph_parser".

jailuthra (approved):
LGTM after squash :ship: :confetti_ball: 

Let's first merge https://github.com/livepeer/FFmpeg/pull/8 and update to the latest commit ID in `install_ffmpeg.sh`

PR #259, opened 2021-09-03
Use static inline instead of inline

cyberj0g:
I ran into 'undefined reference' error when trying to launch debugging from VS Code and GoLand. The reason is that `inline` specifier without `static` not always results in function inlining, and behavior is not consistent across C compilers and C standards. It can be reproduced like that:
````
CGO_CFLAGS="-std=gnu17" go build -gcflags=all="-N -l" cmd/scenedetection/scenedetection.go
````
It's hard to pinpoint exactly which C compiler settings plain `go build` uses, so I'm not sure why it works fine, probably because of C compiler and optimization flags combination. I added `static` specifier to all inline functions, which should produce intended compiler behavior.

### DIFF
--- a/ffmpeg/decoder.h
+++ b/ffmpeg/decoder.h
@@ -57,7 +57,7 @@ int open_audio_decoder(input_params *params, struct input_ctx *ctx);
 void free_input(struct input_ctx *inctx);
 
 // Utility functions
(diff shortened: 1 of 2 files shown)


jailuthra (approved):
Good catch!

PR #262, opened 2021-09-13
Contribution guide - debug build instructions

cyberj0g:


doc/contributing.md
```diff
@@ -1 +1,54 @@
...
+
+To do it:
+1. Copy `~/compiled` dir created by `install_ffmpeg.sh` to `compiled_debug`
+2. If you want to keep system-wide Ffmpeg installation, remove/rename system `pkg-config` (.pc) files of Ffmpeg libs (located in `/usr/local/lib/pkgconfig/` on most Linux distros). Otherwise, `pkg-config` will prioritize system libs without debug information.
+3. [FFmpeg](https://github.com/livepeer/FFmpeg/) should be already cloned after running `install_ffmpeg.sh`. Navigate to FFmpeg dir and run `configure` as below, making sure all paths are correct.
+    ```
+    export ROOT=/projects/livepeer/src
+    export PATH="$ROOT/compiled_debug/bin":$PATH
+    export PKG_CONFIG_PATH="${PKG_CONFIG_PATH:-}:$ROOT/compiled_debug/lib/pkgconfig"
+    ./configure --fatal-warnings \
```
yondonfu, line 18: I wonder if we can incorporate this into the `install_ffmpeg.sh` script. Would be cool if we could do something like `./install_ffmpeg.sh --debug` to do an install with debugging enabled and with the relevant libs installed stored in a dedicated debug directory like `compiled_debug`.

yondonfu (commented):
Awesome! Level up from print statements :)

doc/contributing.md
```diff
@@ -1 +1,54 @@
...
+4. Run `make install`
+5. Make sure there are debug symbols in newly built static libs:
+    ```
+    objdump --syms compiled_debug/lib/libavcodec.a
+    ```
+6. Now you can build Go apps with debug information. Use absolute path in `PKG_CONFIG_PATH` to point to `compiled_debug` dir:
+    ```
+    PKG_CONFIG_PATH=/projects/livepeer/src/compiled_debug/lib/pkgconfig go build -a -gcflags all="-N -l" cmd/scenedetection/scenedetection.go
+    ```
+7. Check that Go executable linked correctly. You shouldn't see any FFmpeg's dynamic libraries in LDD output for newly built binary (e.g. `libavcodec.so`).
```
jailuthra, line 46: Both this and the next one are marked as step # `7`.

It's slightly unclear to me if we're recommending the user to link the `libav*` libraries statically or dynamically.

If they're linked statically - I think `ldd` shouldn't output anything, and in the next step we don't need `LD_LIBRARY_PATH`. We can probably do away with this step altogether if we're going that route. Let me know if I missed something.

doc/contributing.md
jailuthra: +1 on having the debug flags become a part of the `install_ffmpeg.sh` script. We also need to bring this and the `go-livepeer` install ffmpeg script closer to feature parity, with the only difference being explicitly enabling only the protocols, decoders, encoders we need in `go-livepeer` because it's a binary that we release.

But this is a great starting point, and we can handle that in a separate PR.

doc/contributing.md
```diff
@@ -1 +1,54 @@
 # Contributing
+Thank you for interest in contributing to LivePeer ecosystem.
```
jailuthra, line 2:
```suggestion
Thank you for your interest in contributing to Livepeer Media Server.
```

doc/contributing.md
```diff
@@ -1 +1,54 @@
...
+
+# Working on Go code
+If your goal is making changes solely to Go part of the codebase, just follow [Requirements](https://github.com/livepeer/lpms/#requirements) section and make sure you have all dependencies to build Go scripts successfully. Debugging with Delve, GDB, or, visually, with IDE (e.g. JetBrains GoLand, Microsoft VS Code) should work fine.
+
+# Working on CGO interface and FFMpeg code
```
jailuthra, line 7: We can probably add a short note on what works and doesn't with GDB/Delve.

For example with GDB, I've been able to set breakpoints in Go code, LPMS C code by default, and FFmpeg C code with the build steps given here.

But manually `step`ping from Go -> C boundary doesn't work, and `bt` stack traces are also limited to the language boundary.

Feel free to add if it works better with Delve, or an IDE.

jailuthra (approved):
Thanks a *ton* for writing this, it would prove invaluable for new contributors! :smile: 

I made some minor suggestions, but overall content LGTM.

doc/contributing.md
cyberj0g: Shared libraries are more convenient when debugging, updated the doc.

doc/contributing.md
cyberj0g: Done.

PR #274, opened 2021-11-22
add from/to parametrs support for transcoding function

darkdarkdragon:
add from/to parametrs support for transcoding function
from/to specified in ms from segment start

required for [clipping](https://github.com/livepeer/go-livepeer/issues/2072)


cmd/transcoding/transcoding.go
```diff
@@ -42,17 +43,32 @@ func main() {
 	fname := os.Args[1]
 	profiles := str2profs(os.Args[2])
 	accel, lbl := str2accel(os.Args[3])
+	var from, to time.Duration
+	if len(os.Args) > 4 {
```
oscar-davids, line 47: what happened if SW transcoding? So I think it is better to use  flag.Parse().

ffmpeg/api_test.go
```diff
@@ -1288,6 +1288,42 @@ func TestTranscoder_OutputFPS(t *testing.T) {
...
+	out[0].VideoEncoder.Name = "copy"
+	_, err = tc.Transcode(in, out)
+	if err == nil || err != ErrTranscoderClipConfig {
+		t.Errorf("Expected '%s', got %v", ErrTranscoderClipConfig, err)
+	}
+	out[0].From = 0
+	out[0].To = time.Second
+	_, err = tc.Transcode(in, out)
+	if err == nil || err != ErrTranscoderClipConfig {
+		t.Errorf("Expected '%s', got %v", ErrTranscoderClipConfig, err)
```
oscar-davids, line 1316: Is clipping interval 0.0~1.0 an invalid parameter?

oscar-davids (commented):
Overall LGTM! I have some questions and nit for my understanding.

ffmpeg/api_test.go
darkdarkdragon: That test case has set `copy` as encoder - we don't allow clipping for this case.

cmd/transcoding/transcoding.go
```diff
@@ -18,8 +18,9 @@ func validRenditions() []string {
 }
 
 func main() {
+	var err error
 	if len(os.Args) <= 3 {
-		panic("Usage: <input file> <output renditions, comma separated> <sw/nv>")
+		panic("Usage: <input file> <output renditions, comma separated> <sw/nv> <from> <to>")
```
jailuthra, line 23: I think there's a missing argument here: 
```suggestion
		panic("Usage: <input file> <output renditions, comma separated> <sw/nv> [<gpu_id>] <from> <to>")
```
So I agree with @oscar-davids that using [flag](https://pkg.go.dev/flag) for parsing might be a good idea here, and have consistent argument names compared to the `livepeer` and `livepeer_bench` tools.

ffmpeg/api_test.go
```diff
@@ -1288,6 +1288,42 @@ func TestTranscoder_OutputFPS(t *testing.T) {
 	outputFPS(t, Software)
 }
 
+func TestTranscoderAPI_ClipInvalidConfig(t *testing.T) {
+	// Test the following file open results on input: fail, success, fail, success
```
jailuthra, line 1292: Are we missing the success test cases here?

Also given clipping touches our timestamp logic, which is different for CPU and GPU, it might be a good idea to have this test case inside a common `clipConfig()` function that is called both for CPU here, and for GPU devices in `nvidia_test.go`.

jailuthra (commented):
Functionally LGTM. Made some minor comments around tests and tooling improvement.

cmd/transcoding/transcoding.go
darkdarkdragon: fixed in 9b838127744b93d928de8c9294267caaabb3e599

ffmpeg/api_test.go
darkdarkdragon: > Are we missing the success test cases here?

It was wrong comment, removed.

> , which is different for CPU and GPU

Can you elaborate on that?

ffmpeg/ffmpeg.go
```diff
@@ -269,6 +273,16 @@ func (t *Transcoder) Transcode(input *TranscodeOptionsIn, ps []TranscodeOptions)
 	if err != nil {
 		return nil, err
 	}
+	for _, p := range ps {
+		if (p.From > 0 || p.To > 0) && (p.VideoEncoder.Name == "drop" || p.VideoEncoder.Name == "copy") {
```
oscar-davids, line 277:
```suggestion
		if (p.To < p.From || p.From < 0){
		        glog.Warning("'To' should be after 'From' and should be positive integer")
			return nil, ErrTranscoderClipConfig
		}
		if (p.VideoEncoder.Name == "drop" || p.VideoEncoder.Name == "copy") {
		        glog.Warning("Could clip only when transcoding video")
			return nil, ErrTranscoderClipConfig
```
I think the above suggestion works as well because the following parameters could be passed under the here conditions.
  from -1 to -2, from 0 to -3, from -3 to 2

ffmpeg/api_test.go
jailuthra: > Can you elaborate on that?

My bad, the timestamps should match between CPU and GPU pipelines. The difference only occurs if FPS filter (and our timestamp logic around it) is used. So we can keep a single test case here.

ffmpeg/ffmpeg.go
darkdarkdragon: thanks for catching that up!
fixed in fc2eb738651f2238bd98c48f4de73004e8c6b342

PR #275, opened 2021-11-29
FFmpeg Quirks in LPMS

oscar-davids:
This document outlines how LPMS tweaks FFmpeg transcoding pipeline to address some specific problems.


QUIRKS.md
```diff
@@ -0,0 +1,67 @@
...
+                                                    |
+                                                    |
+                                                    +
+                                           (1) dummy monotonic
+                                                   pts
+```
+
+FPS filter expects monotonic increase in input frame's PTS. We cannot rely on the input to be monotonic thus:
+i. Set a dummy PTS before the frame is sent into filtergraph, that we manually increase monotonically.
+ii. OR use SETPTS filter in the filtergraph before FPS filter, which would do the same thing.
```
jailuthra, line 58: We should remove this from the documentation, as we did not take the approach of SETPTS filter due to reasons mentioned here https://github.com/livepeer/lpms/pull/201#issuecomment-664853222

QUIRKS.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Very-few-video-frame segment handling by introducing sentinel frames
+
+### Problem
+Hardware transcoding fails when livepeer is sent segments with very few video frames. It works fine when running software trascoding but fails in hardware trascoding. This is caused because internal buffers of Nvidia buffers are bigger than in software mode. This could have been addressed by using ffmpeg flushing at the end of each segment however, when we do ffmpeg flushing we need to close and reopen transcoding session and this is quite expensive in Nvidia. Instead of flushing, closing and reopening transcoding session for each segment, LPMS chose to reuse the session for different segments of the same stream.
+
+### Solution
+To solve the flushing problem while still reusing the session, we introduced so called sentinel-packets. Sentinel packets are dummy frames with -1 timestamps. We insert these packets at the end of each segment to make sure that the packets that are sent to the buffer earlier always get popped out. We wait until we receive the sentinel packets pack and if we receive sentinel packet we know that this is the end of the segment.
+
```
jailuthra, line 22: We should probably add links to the code snippet that implements this.
We can also note the fact that the filter gives up after sending `SENTINEL_MAX` packets, which is a pre-processor constant defined as `5` for now.

QUIRKS.md
```diff
@@ -0,0 +1,67 @@
...
+## Handle zero frame (audio-only) segments at the start of a session
+
+### Problem
+Livepeer used to reject transcoding requests when it is sent audio-only segments. This could happen variety of reasons. In the case of MistServer, it happens when a transcode is enabled and there is more audio data in the buffer than video data. It could also happen in a variety of edge cases where bandwidth is severely constrained and end-user client software decides to only send audio to keep something online. 
+
+### Desired Behavior
+Instead of erroring on such segments, let's accept them and send back audio-only tracks. The audio output should be exactly the same as the audio input.
+
+### Our Solution
+We bypass the usual transcoding process for audio-only segments and just return them back. To do this, we check if video frame is present, and if not present just copy the frames to the output as it is. The bypassing is done by forcing the copy transcoder https://github.com/livepeer/lpms/commit/8bc28e3f702049a17c24ab2041857a47d8af51bf for such segments.
```
jailuthra, line 13: While the by-pass check code is still implemented in LPMS we no longer use this function on the transcoder directly, or set the 'copy' encoder. https://github.com/livepeer/lpms/blob/46f7030c62d91d5f74b02bfa13dca42eb8ad8fa5/ffmpeg/extras.c#L110-L117 

We instead moved the calling logic on the Broadcaster side in https://github.com/livepeer/go-livepeer/pull/1933 and the document should reflect that.

jailuthra (changes requested):
Would be good to have links and refrences to issues and PRs around these topics, as a time-saver for engineers who on-board later on.

cyberj0g:
Amazing write-up, @jailuthra @oscar-davids! Hijacking this branch to add a small section on reusing transcoding session with HW codecs. I believe much of the information on this is already in the document.

PR #276, opened 2021-12-03
added fail-case unit test

oscar-davids:
Add test suite according to [this](https://github.com/livepeer/internal-project-tracking/issues/248) issue.
- first, download with following command
 `gsutil cp -r gs://livepeer-production-failed-transcodes /home/livepeer-production-failed-transcodes`
- enter transcoder directory
 `cd transcoder `
- run unit test with following command
 `FAILCASE_PATH="/home/livepeer-production-failed-transcodes" go test --tags=nvidia -timeout 30m -run Nvidia`
 It took about 15 minutes on my side.
- can double-check as result.csv file in the same directory


transcoder/ffmpeg_segment_failcase_test.go
```diff
@@ -0,0 +1,266 @@
...
+		t.Error("Failed in parsing .txt based files.")
+	}
+	fmt.Println("Start transcoding from failed source files.")
+
+	failcount, lasterr := checkTranscodingFailCase(t, inTparams, accel, outdir, outcsv)
+
+	if lasterr != nil {
+		t.Error("Failed in checking fail case files.")
+	}
+	fmt.Println("Succeed in fail check- real failed:", failcount, " / total:", len(inTparams))
```
cyberj0g, line 265: Could we have a more concise message, like: "Test completed, really failed: XXX / total: XXX"

cyberj0g (approved):
Thanks @oscar-davids, added a small remark on log message, otherwise LGTM, tested locally. A separate thing, we probably won't have any info on output codec requested (and, potentially, other capabilities) in `target_profiles` of JSON files of the bucket, so with multiple codecs we won't actually know what exactly failed, maybe create an issue to update this structure? This test will probably need to be updated too after that. @yondonfu @jailuthra

transcoder/ffmpeg_segment_failcase_test.go
```diff
@@ -0,0 +1,266 @@
...
+func getTsFilename(txtpath string) string {
+	txtfname := path.Base(txtpath)
+	tsfname := ""
+	i := strings.Index(txtfname, ".ts-")
+	if i > 0 {
+		tsfname = txtfname[:i] + ".ts"
+	}
+	return tsfname
+}
+func parseProfiles(instr string) []ffmpeg.VideoProfile {
```
jailuthra, line 44: Given this is used at multiple places (livepeer_bench another example), maybe we should move this some place common, like `ffmpeg/videoprofile.go`, to avoid duplication?

jailuthra (approved):
Overall LGTM. Made a minor comment about having a common JSON -> ffmpeg.VideoProfile parser.

> A separate thing, we probably won't have any info on output codec requested (and, potentially, other capabilities) in target_profiles of JSON files of the bucket, 

Hmm good point, maybe we should store the complete JSON that is returned via the stream-auth-webhook instead of just the target profiles. https://github.com/livepeer/go-livepeer/blob/7af7b79fc0c859f67f0a9318ba5549d7da4b16f6/server/mediaserver.go#L105

That should contain any other user configuration for the failing stream/segment, including detection, verification or output codecs in future.

transcoder/ffmpeg_segment_failcase_test.go
oscar-davids: Good Idea, then we also need to move the following functions.

DefaultProfileName,
EncoderProfileNameToValue

And last one thing,  we have to refactor go-livepeer part related to these functions. WDYT?

jailuthra: >  then we also need to move the following functions.
DefaultProfileName,
EncoderProfileNameToValue
And last one thing, we have to refactor go-livepeer part related to these functions. 

Yeah makes sense on moving these functions from `go-livepeer/common` to `lpms/ffmpeg` as well. Feel free to do this refactoring separately from this PR if you think that is cleaner.

yondonfu:
Looks like there is some additional information that would be useful to track for segments that failed transcoding.

@oscar-davids Can you create an issue to track this?

Issue #241, opened 2021-06-23
ffmpeg/api_test: Fix commented-out unit tests


jailuthra:
We've got some commented-out unit tests that we should uncomment or move to a separate file with an `optional` build tag similar to what we've done for `nvidia` specific unit tests.
https://github.com/livepeer/lpms/blob/301900dd25100b0624104233c5ad644d9b023960/ffmpeg/api_test.go#L1297

yondonfu:
Closed by #277

PR #277, opened 2021-12-03
Uncomment the codes for the NoKeyframe specific unit test

JamesWanglf:
This PR is related with [this issue](https://github.com/livepeer/lpms/issues/241). I have uncomment the codes and uploaded two ts files for the test, **kryp-1.ts** is the one that contains keyframes, while **kryp-2.ts** has no keyframe.


ffmpeg/api_test.go
```diff
@@ -1322,6 +1322,46 @@ func TestTranscoderAPI_ClipInvalidConfig(t *testing.T) {
 	}
 }
 
+func noKeyframeSegment(t *testing.T, accel Acceleration) {
+	// Reproducing #219
+	run, dir := setupTest(t)
+	defer os.RemoveAll(dir)
+
+	cmd := `
+		cp "$1"/../transcoder/kryp-*.ts .
```
jailuthra, line 1331: We should move these segments to `data` instead of `transcoder` directory, and update this line, as we now store all test segments there.

jailuthra (commented):
LGTM, and tested on my GTX 1060. Made a minor comment about moving the segments to `data/` directory instead.

There is an unexplored possibility of generating a segment without keyframe packets on-the-fly using ffmpeg commands, but I think the current solution is much cleaner so we can stick to it.

cyberj0g (approved):
LGTM!

oscar-davids (approved):
LGTM!

jailuthra (approved):
LGTM :ship:

PR #279, opened 2021-12-09
Support H.265/HEVC, VP8, VP9

cyberj0g:
Issue link: https://github.com/livepeer/go-livepeer/issues/2123

Details:
* Added support for H.265 codecs: libx265 (CPU, encoding and decoding), NVENC (GPU, encoding and decoding)
* Added support for VP8/VP9 codecs: libvpx (CPU, encoding and decoding), NVENC (GPU, decoding only, where supported)
* Added transcoding test for all combinations of codecs and acceleration

Testing:
* install dependencies using [install_ffmpeg.sh](https://github.com/livepeer/go-livepeer/blob/ip/hevc/install_ffmpeg.sh) (WIP, please report errors)
* run tests:
  ```
  go test -v --tags=nvidia -timeout 30m -run TestNvidia_Transcoding ./ffmpeg/
  go test -v --tags=nvidia -timeout 30m -run TestSW_Transcoding ./ffmpeg/
  ```
They will output lines like below, and fail, if anything is not supported (on CI, only CPU test will run for now). SW test should work anywhere, Nvidia test should pass on GPUs starting from desktop GTX 1070 - it can't decode VP9 on my GTX 1070m, despite support matrix stating it should.
```
    ffmpeg_test.go:754: Transcode VP8 (Accel: 0) -> H.265/HEVC (Accel: 0) Prepare: true Transcode: true Quality: true
    ffmpeg_test.go:754: Transcode VP8 (Accel: 0) -> VP8 (Accel: 0) Prepare: true Transcode: true Quality: true
    ffmpeg_test.go:754: Transcode VP8 (Accel: 0) -> VP9 (Accel: 0) Prepare: true Transcode: true Quality: true
```


cyberj0g:
Finished testing new 'unified' [install_ffmpeg.sh](https://github.com/livepeer/go-livepeer/blob/ip/hevc/install_ffmpeg.sh) from go-livepeer and added it to circle-ci job. We shouldn't need a separate one for lpms now. For dev/test install, add `debug-video` to BUILD_TAGS e.g.:
```
BUILD_TAGS="experimental" ./install_ffmpeg.sh /projects/livepeer/src/
```
I checked new codecs tests with this `install_ffmpeg.sh`, but not sure if all other tests are fine. @JamesWanglf is welcome to add his changes and create PR in go-livepeer.

leszko:
> @leszko What is the reason for including `matroska,opus,webm,webm_chunk,webm_dash_manifest` into production build?

hmm? What do you mean?

darkdarkdragon:
> > @leszko What is the reason for including `matroska,opus,webm,webm_chunk,webm_dash_manifest` into production build?
> 
> hmm? What do you mean?

In install_ffmpeg.sh you enabled those muxers by default, which means that they will go into production `go-livepeer` build.

darkdarkdragon:
> > > > @leszko What is the reason for including `matroska,opus,webm,webm_chunk,webm_dash_manifest` into production build?
> > > 
> > > 
> > > hmm? What do you mean?
> > 
> > 
> > In install_ffmpeg.sh you enabled those muxers by default, which means that they will go into production `go-livepeer` build.
> 
> Did I? Where is this change?

Sorry, it was not you 😄

cyberj0g:
@darkdarkdragon I added these. It makes sense to support MKV and WebM containers for VP8/9, and AAC is not supported for WebM - that's why Opus is added. DASH is a great alternative to HLS (especially for WebM), we might want to support it in the near future too.

leszko:
> > > > > @leszko What is the reason for including `matroska,opus,webm,webm_chunk,webm_dash_manifest` into production build?
> > > > 
> > > > 
> > > > hmm? What do you mean?
> > > 
> > > 
> > > In install_ffmpeg.sh you enabled those muxers by default, which means that they will go into production `go-livepeer` build.
> > 
> > 
> > Did I? Where is this change?
> 
> Sorry, it was not you 😄

🙃

darkdarkdragon:
> @darkdarkdragon I added these. It makes sense to support MKV and WebM containers for VP8/9, and AAC is not supported for WebM - that's why Opus is added. DASH is a great alternative to HLS (especially for WebM), we might want to support it in the near future too.

Any addition to the production build increases attack surface - so we shouldn't add anything that don't officially support. 
So I think we should remove `webm_dash_manifest` and `matroska`, and create new issue in `go-livepeer` repo for adding support for `.webm` segments.

Also, shouldn't `--enable-libopus` be added for `opus` support?

cyberj0g:
> Any addition to the production build increases attack surface - so we shouldn't add anything that don't officially support. 

That's a good practice, but aren't we adding VP8/9 support at least for decoding with this PR? If so, common containers used with these are MKV and WEBM, and lpms now supports them. Or, since VP8/9 is an exotic choice for HLS+MPEG-TS, we should not enable these codecs at all until other streaming protocols are not implemented on go-livepeer side?

darkdarkdragon:
> 

I think `webm` would be enough? But it's better to create issue in go-livepeer to discuss what containers should we support.

> Or, since VP8/9 is an exotic choice for HLS+MPEG-TS, we should not enable these codecs at all until other streaming protocols are not implemented on go-livepeer side?

We're really not using HLS output of go-livepeer node nowdays - it is being handled by Mist (I think in the near future, when we will integrate Mist and go-livepeer better we'll remove RTMP and HLS support from go-livepeer node)
So it is enough for go-livepeer node to being able to take .webm file and respond back with .mp4 or .ts (or .webm).

cyberj0g:
Created a [new issue](https://github.com/livepeer/lpms/pull/279) for additional container formats, and moved these flags to debug build for now.

oscar-davids (commented):
Overall LGTM!

As a separate matter, one thing I'm thinking of is to statically link x265 and libvpx like x264. Because if we link it dynamically, we need to set LD_LIBRARY_PATH or add related .so files to the distribution.  WDYT?

and I wonder how this [part ](https://github.com/livepeer/lpms/blob/0999e7e2f16c2065a6f6d2eb5b94411bfe911656/ffmpeg/ffmpeg.go#L398)will work on the VP8/VP9/H265.

cyberj0g:
> one thing I'm thinking of is to statically link x265 and libvpx like x264

Spent quite a bit of time and got Ffmpeg to [build with new codecs as static libraries](https://github.com/livepeer/go-livepeer/blob/ip/new-codecs-static/install_ffmpeg.sh). However, libx265 is a C++ lib, and I couldn't get CGO to build lpms properly when it statically linked, and don't know if it's a good idea to switch to g++ in lpms. At this point, I believe it's not worth the effort to get it to build statically with libx265. If someone is willing to continue looking into that, my efforts stopped there (set proper PKG_CONFIG_PATH and use linked `install_ffmpeg.sh`):
```
CGO_CFLAGS="-Wno-error" CGO_CPPFLAGS="-Wno-error" CC=g++ go build -a cmd/transcoding/transcoding.go
# runtime/cgo
In file included from gcc_context.c:8:
libcgo.h:97:15: error: unnecessary parentheses in declaration of ‘_cgo_get_context_function’ [-Werror=parentheses]
   97 | extern void (*(_cgo_get_context_function(void)))(struct context_arg*);
      |               ^
cc1plus: all warnings being treated as errors
```
Without `CC=g++`, I'm getting:
```
/usr/bin/ld: /projects/livepeer/src/compiled/lib/libx265.a(encoder.cpp.o): undefined reference to symbol '_ZTVN10__cxxabiv121__vmi_class_type_infoE@@CXXABI_1.3'
/usr/bin/ld: /usr/lib/x86_64-linux-gnu//libstdc++.so.6: error adding symbols: DSO missing from command line
```

cyberj0g:
> and I wonder how this part will work on the VP8/VP9/H265.

It doesn't, none of these parameters work for HEVC or VP8/9 to my knowledge.

iameli:
@cyberj0g Does the node boot without libx265, only loading it dynamically when needed? I don't think we probably want to add an external libx265 dependency if not.

cyberj0g:
> @cyberj0g Does the node boot without libx265, only loading it dynamically when needed? I don't think we probably want to add an external libx265 dependency if not.

I think we agreed to not include software versions of new codecs into production build for now, so it won't be an issue. In debug mode, `--enable-shared` is added, so all dependencies are linked dynamically. As an experiment, I quickly tried building without `--enable-shared` with all dependencies static, except `libx265.so` - it didn't work, probably something is missing from Ffmpeg's `configure.sh` to support mixing static and shared dependencies.

ffmpeg/ffmpeg_test.go
```diff
@@ -28,12 +28,14 @@ func setupTest(t *testing.T) (func(cmd string), string) {
 	// Executes the given bash script and checks the results.
 	// The script is passed two arguments:
 	// a tempdir and the current working directory.
-	cmdFunc := func(cmd string) {
+	cmdFunc := func(cmd string) bool {
```
yondonfu, line 31: Do we need to return a bool here if the cmd string executed via `exec.Command()` triggers an error if it ends with an exit code of 1?

See [this test example](https://github.com/livepeer/lpms/blob/5c05cbbeb53bf7cfbc413c5b33ddadf3d3699719/ffmpeg/ffmpeg_test.go#L107) that I believes will cause the test to fail if the bash command using ffprobe has an exit code of 1.

ffmpeg/ffmpeg_test.go
```diff
@@ -602,6 +604,159 @@ func TestTranscoder_MuxerOpts(t *testing.T) {
...
+			}, []TranscodeOptions{
+				{
+					Oname:   inName,
+					Profile: profile,
+					Accel:   Software,
+				},
+			})
+			if err != nil {
+				t.Error(err)
+				t.Fail()
```
yondonfu, line 696: [t.Error](https://pkg.go.dev/testing#T.Error) is equivalent to `t.Log` followed by `t.Fail` so the `t.Fall` call should not be necessary.

ffmpeg/ffmpeg_test.go
```diff
@@ -602,6 +604,159 @@ func TestTranscoder_MuxerOpts(t *testing.T) {
 	run(cmd)
 }
 
+type TranscodeOptionsTest struct {
+	InputCodec   VideoCodec
+	OutputCodec  VideoCodec
+	InputAccel   Acceleration
+	OutputAccel  Acceleration
+	Profile      VideoProfile
+	QualityCheck bool
```
yondonfu, line 613: Is this ever not true?

ffmpeg/ffmpeg_test.go
```diff
@@ -602,6 +604,159 @@ func TestTranscoder_MuxerOpts(t *testing.T) {
...
+						QualityCheck: true,
+					})
+				}
+			}
+		}
+	}
+	return opts
+}
+
+func CodecsComboTest(t *testing.T, options []TranscodeOptionsTest) {
```
yondonfu, line 652: This probably doesn't need to be exported i.e. could be `codecsComboTest`.

ffmpeg/ffmpeg_test.go
```diff
@@ -602,6 +604,159 @@ func TestTranscoder_MuxerOpts(t *testing.T) {
...
+			cmd = fmt.Sprintf(`
+    # compare using ssim and generate stats file
+    ffmpeg -loglevel warning -i %s -i %s -lavfi '[0:v][1:v]ssim=stats.log' -f null -
+    # check image quality; ensure that no more than 5 frames have ssim < 0.95
+    grep -Po 'All:\K\d+.\d+' stats.log | awk '{ if ($1 < 0.95) count=count+1 } END{ exit count > 5 }'
+  `, outName, qName)
+			if quality {
+				quality = run(cmd)
+				if !quality {
+					t.Fail()
```
yondonfu, line 747: Shouldn't the cmd trigger a `t.Error()` if it returns a non-zero exit code?

yondonfu (changes requested):

ffmpeg/decoder.c
```diff
@@ -211,24 +211,40 @@ int open_audio_decoder(input_params *params, struct input_ctx *ctx)
...
   else if (ctx->vi < 0) {
     LPMS_WARN("No video stream found in input");
   } else {
     if (AV_HWDEVICE_TYPE_CUDA == params->hw_type) {
-      if (AV_CODEC_ID_H264 != codec->id) {
+      char* decoder_name = get_hw_decoder(codec->id);
+      if (!*decoder_name) {
         ret = lpms_ERR_INPUT_CODEC;
-        LPMS_ERR(open_decoder_err, "Non H264 codec detected in input");
+        LPMS_ERR(open_decoder_err, "Input codec does not support hardware acceleration");
```
yondonfu, line 245: WDYT about adding a unit test for this i.e. passing in a non-supported codec for the decoder?

ffmpeg/ffmpeg_test.go
cyberj0g: I think the func produced by `setupTest` never checks exit code, it will trigger error if [something was written to stderr](https://github.com/livepeer/lpms/blob/5c05cbbeb53bf7cfbc413c5b33ddadf3d3699719/ffmpeg/ffmpeg_test.go#L33), which won't be the case here.

cyberj0g: Fixed.

ffmpeg/ffmpeg_test.go
cyberj0g: Removed.

cyberj0g (commented):
Thanks for review @yondonfu, I addressed all comments.

ffmpeg/decoder.c
yondonfu: Ah I see. Noticed [this comment](https://github.com/livepeer/lpms/blob/8f053cf92946f36cb2873a896b0db06713125142/ffmpeg/nvidia_test.go#L17) that should be updated then.

ffmpeg/ffmpeg_test.go
yondonfu: Ah right so `t.Error()` will log the error, but if the caller of the command wants to know if the command failed it currently doesn't have a way to know that and returning a bool solves that. Got it.

yondonfu (approved):
LGTM! I'd like to get an approval from another set of eyes that has familiarity with this codebase as well. cc @oscar-davids since you already left an initial review.

I left a minor comment about updating a comment - feel free to address prior to merging. Also, I recommend rebasing off master to pick up a fix that landed recently that should address the failing test in CI and to squash the commits to clean up the commit history before merging.

ffmpeg/encoder.c
```diff
@@ -232,7 +232,7 @@ int open_output(struct output_ctx *octx, struct input_ctx *ictx)
     if (octx->fps.den) vc->time_base = av_buffersink_get_time_base(octx->vf.sink_ctx);
     else if (ictx->vc->time_base.num && ictx->vc->time_base.den) vc->time_base = ictx->vc->time_base;
     else vc->time_base = ictx->ic->streams[ictx->vi]->time_base;
-    if (octx->bitrate) vc->rc_min_rate = vc->rc_max_rate = vc->rc_buffer_size = octx->bitrate;
+    if (octx->bitrate) vc->rc_min_rate = vc->bit_rate = vc->rc_max_rate = vc->rc_buffer_size = octx->bitrate;
```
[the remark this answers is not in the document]
yondonfu: Gotcha. Seems reasonable, but I'm curious if this has resulted in any issues for anyone.

cc @iameli - have we seen issues with bitrate before?

darkdarkdragon (approved):
LGTM

oscar-davids (approved):
LGTM!
please rebase on master before merging.

Issue #278, opened 2021-12-07
CircleCI failing on `segmenter` tests


jailuthra:
Since last week our CI pipeline fails on `TestSegmenter` unit test here https://github.com/livepeer/lpms/blob/1b6c4df523c2523d5309cd0fe9e7c518811ea779/segmenter/video_segmenter_test.go#L134-L138

Nothing obvious has changed on the segmenter codebase recently, so I suspect something changed in the environment (i.e. CircleCI, Golang, RTMP port?) that is causing this.

yondonfu:
Let's determine if this is a problem with CircleCI vs. a problem with the tests/code.

If the problem is with the tests/code let's fix it.

If the problem is with CircleCI I suggest we look into https://github.com/livepeer/lpms/issues/284

yondonfu:
Closed by #286

PR #281, opened 2021-12-16
fix non-monotonic increase audio frames issue

oscar-davids:
A follow up to[ #249](https://github.com/livepeer/internal-project-tracking/issues/249).

**Problem:**

Transcoding pipeline would fail when audio frames with non-monotonic increase DTS were sent to "mux" filter. To be more precise, in the [fail-case test](https://github.com/livepeer/internal-project-tracking/issues/249), the transcoding pipeline returned "Invalid argument" error when the order of the audio AVpacket's DTS is duplicated or smaller than the previous one.
The error was solved by making the monotonic audio AVpacket's DTS like the [pipeline ](https://github.com/livepeer/FFmpeg/blob/682c4189d8364867bcc49f9749e04b27dc37cded/fftools/ffmpeg.c#L824)of ffmpeg.

**Testing:**

`cd transcoder`
`FAILCASE_PATH="/home/livepeer-production-failed-transcodes" go test --tags=nvidia -timeout 30m -run Nvidia`

we can see following message
`Test completed, really failed: 650  / total: 1386`
261 "Invalid argument" transcoding errors have been resolved.

or 

`go test -run TestTranscoder_NonMonotonicAudioSegment ./ffmpeg`
`go test --tags=nvidia -run TestNvidia_NonMonotonicAudioSegment ./ffmpeg`


yondonfu (commented):
Interesting find! A few questions:

- Sounds like the out-of-order audio frames triggers the "Invalid argument" error during muxing so this is a problem for both software & Nvidia transcoding right?
- Are there any more informative error logs preceding the "Invalid argument" error when this out-of-order audio frames issue occurs?

Regarding testing: Great to see that the "Invalid argument" errors from the fail case test are fixed after these changes. What do you think about also adding a unit test using a segment with out-of-order audio frames that can be used to reproduce this specific scenario similar to the unit tests that we have for other bugs that needed to be fixed in the past (i.e. no keyframe, single keyframe, etc.)? Having this unit test in our test suite would make it clearer to everyone going through the codebase that the transcoding pipeline should be able to handle out-of-order audio frames and ensures that it continues to handle them even as additional changes are made in the future.

yondonfu:
I'm also curious about the scenarios where we would expect to receive segments with out-of-order audio frames. 

Since the segments with out-of-order audio frames that were used in the fail case test came from Mist, I wonder if this is just expected to happen based on a streamer's setup from time to time or if there is something unexpected happening on the Mist side.

@Thulinma any thoughts on this?

oscar-davids:
> * Sounds like the out-of-order audio frames triggers the "Invalid argument" error during muxing so this is a problem for both software & Nvidia transcoding right?

Right,  the out-of-order audio frames trigger an "Invalid argument" error in both software and Nvidia transcoding.

> * Are there any more informative error logs preceding the "Invalid argument" error when this out-of-order audio frames issue occurs?

There is no other informative error log other than the "invalid argument" error. Here is a snippet for one segment.
```
[Parsed_fps_1 @ 0x2130bc0] Writing frame with pts 1 to pts 1
[mpegts @ 0x2034300] Application provided invalid, non monotonically increasing dts to muxer in stream 1: 185047248 >= 185047248
ERROR: transcoder.c:316] Error encoding : Invalid argument
[AVIOContext @ 0x202f080] Statistics: 1167476 bytes read, 2 seeks
[AVIOContext @ 0x2044180] Statistics: 0 seeks, 0 writeouts
ERROR: logging before flag.Parse: E1216 11:09:22.342118    8350 ffmpeg.go:522] Transcoder Return : Invalid argument
```

> Regarding testing: Great to see that the "Invalid argument" errors from the fail case test are fixed after these changes. What do you think about also adding a unit test using a segment with out-of-order audio frames that can be used to reproduce this specific scenario similar to the unit tests that we have for other bugs that needed to be fixed in the past (i.e. no keyframe, single keyframe, etc.)? Having this unit test in our test suite would make it clearer to everyone going through the codebase that the transcoding pipeline should be able to handle out-of-order audio frames and ensures that it continues to handle them even as additional changes are made in the future.

I totally agree, That makes sense. I will add unit tests like past no key frame case.

Thulinma:
> I'm also curious about the scenarios where we would expect to receive segments with out-of-order audio frames.
> 
> Since the segments with out-of-order audio frames that were used in the fail case test came from Mist, I wonder if this is just expected to happen based on a streamer's setup from time to time or if there is something unexpected happening on the Mist side.
> 
> @Thulinma any thoughts on this?

Wait... what? This is quite worrying to me, as Mist's backend is not _capable_ of producing out-of-order packets 🤔 It _can_ duplicate timestamps - in theory - but it would only do so if the input timestamps have that same duplication already going into it. Do we have any more information on how this happened / where the stream came from?

ffmpeg/encoder.c
```diff
@@ -376,6 +379,20 @@ int mux(AVPacket *pkt, AVRational tb, struct output_ctx *octx, AVStream *ost)
...
       if (octx->drop_ts == AV_NOPTS_VALUE) octx->drop_ts = pkt->pts;
       if (pkt->pts && pkt->pts == octx->drop_ts) return 0;
+
+      if (pkt->dts != AV_NOPTS_VALUE && pkt->pts != AV_NOPTS_VALUE && pkt->dts > pkt->pts) {
+        pkt->pts = pkt->dts = pkt->pts + pkt->dts + octx->last_audio_dts + 1
+                     - FFMIN3(pkt->pts, pkt->dts, octx->last_audio_dts + 1)
+                     - FFMAX3(pkt->pts, pkt->dts, octx->last_audio_dts + 1);
+      }
+      if (pkt->dts != AV_NOPTS_VALUE && octx->last_audio_dts != AV_NOPTS_VALUE) {
+        int64_t max = octx->last_audio_dts + !(octx->oc->oformat->flags & AVFMT_TS_NONSTRICT);
```
cyberj0g, line 389: Looks like `last_audio_dts` used before initialization here? That's a bit esoteric snippet, maybe add a comment on what it does?

oscar-davids: Added detail comments for last_audio_dts variable, and this pipeline is ported from [here](https://github.com/livepeer/FFmpeg/blob/682c4189d8364867bcc49f9749e04b27dc37cded/fftools/ffmpeg.c#L824)

oscar-davids:
> Wait... what? This is quite worrying to me, as Mist's backend is not capable of producing out-of-order packets 🤔 It can duplicate timestamps - in theory - but it would only do so if the input timestamps have that same duplication already going into it. Do we have any more information on how this happened / where the stream came from?

@Thulinma I added test data [here](https://github.com/livepeer/lpms/blob/oc/audiodts/data/audio-dts.ts), which has duplicate audio timestamps.

cyberj0g:
LGTM, great work @oscar-davids!

cyberj0g (approved):

Thulinma:
> @Thulinma I added test data [here](https://github.com/livepeer/lpms/blob/oc/audiodts/data/audio-dts.ts), which has duplicate audio timestamps.

Thanks. It looks like there's a _bunch_ of duplicate timestamps in there, but thankfully no out-of-order ones. Interesting. 🤔
My best guess right now is to assume these duplicates were already in the input stream, and Mist preserved them as-is. After all, there is not much else it could do in that situation, without compromising the integrity of the stream itself...

ffmpeg/encoder.c
```diff
@@ -376,6 +379,23 @@ int mux(AVPacket *pkt, AVRational tb, struct output_ctx *octx, AVStream *ost)
   if (AVMEDIA_TYPE_AUDIO == ost->codecpar->codec_type) {
       if (octx->drop_ts == AV_NOPTS_VALUE) octx->drop_ts = pkt->pts;
       if (pkt->pts && pkt->pts == octx->drop_ts) return 0;
+
+      if (pkt->dts != AV_NOPTS_VALUE && pkt->pts != AV_NOPTS_VALUE && pkt->dts > pkt->pts) {
```
yondonfu, line 383: WDYT about adding a permalink to the relevant code from the ffmpeg CLI that this is taken from as a reference?

yondonfu (approved):
LGTM! A few comments that you can feel free to address before merging:

- Could a more descriptive name for `audio-dts.ts` be `duplicate-audio-dts.ts`?
- Why did the file permissions change for the `sign_*` files?

I also left a comment about adding a permalink to the ffmpeg CLI code reference.

yondonfu:
Let's try to get the CI tests fixed first though and merge when CI is green. Related https://github.com/livepeer/lpms/issues/278

ffmpeg/encoder.c
oscar-davids: > Could a more descriptive name for audio-dts.ts be duplicate-audio-dts.ts?
> Why did the file permissions change for the sign_* files?
> WDYT about adding a permalink to the relevant code from the ffmpeg CLI that this is taken from as a reference?

Changed audio-dts.ts to duplicate-audio-dts.ts.
There is no special reason, just remove the execute permissions for those files.
Added permalink to the relevant code from the ffmpeg CLI.

yondonfu (approved):
LGTM

Bump on this question:

> Why did the file permissions change for the sign_* files?

oscar-davids:
> Why did the file permissions change for the sign_* files?

There is no special reason, just removed the execute permissions for those files.

PR #286, opened 2021-12-28
fix segmenter test

darkdarkdragon:
Turned out when test function pumps data to TCP connection too fast it leads to that connection being reset.
Add small delay to that test function to slow down data transfer and that fixed issue.


oscar-davids (approved):
LGTM!

Issue #271, opened 2021-11-05
Changing pixel format and/or codec mid-stream triggering CUDA OOM


yondonfu:
See https://discord.com/channels/423160867534929930/731273721125273655/904906439695413288

yondonfu:
For now, we'll focus on returning a more informative error message for this case.

yondonfu:
Closed by #287

PR #287, opened 2021-12-30
fix middle pixel format changing issue

oscar-davids:
A follow up to #271.

In HW transcoding, demuxer is opened once and used, so it is necessary to check whether the input pixel format does not change in the middle. If a change is detected, free the input_ctx and reopen the demuxer again.

test 

```
 go test -run TestTranscoder_DiscontinuityPixelFormat ./ffmpeg
 go test --tags=nvidia -run TestNvidia_DiscontinuityPixelFormat ./ffmpeg
```

### DIFF
--- a/ffmpeg/transcoder.c
+++ b/ffmpeg/transcoder.c
@@ -152,6 +152,7 @@ int transcode(struct transcode_thread *h,
       free_input(&h->ictx);
       ret = open_input(inp, &h->ictx);
       if (ret < 0) LPMS_ERR(transcode_cleanup, "Unable to reopen video demuxer for HW decoding");
+      reopen_decoders = 0;
     }
   }
 
--- a/ffmpeg/transcoder.c
+++ b/ffmpeg/transcoder.c
@@ -148,7 +148,10 @@ int transcode(struct transcode_thread *h,
     else if (ictx->ic->streams[ictx->vi]->codecpar->format != ictx->last_format) {
       LPMS_WARN("Input pixel format has been changed in the middle.");
       ictx->last_format = ictx->ic->streams[ictx->vi]->codecpar->format;
-      // free and reopen demuxer
+      // if the decoder is not re-opened when the video pixel format is changed,
+      // the decoder tries HW decoding with the video context initialized to a pixel format different from the input one.
+      // to handle a change in the input pixel format,
+      // we close the demuxer and re-open the decoder by calling open_input().
       free_input(&h->ictx);
       ret = open_input(inp, &h->ictx);
       if (ret < 0) LPMS_ERR(transcode_cleanup, "Unable to reopen video demuxer for HW decoding");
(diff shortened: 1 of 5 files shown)


cyberj0g:
LGTM!

cyberj0g (approved):

ffmpeg/transcoder.c
```diff
@@ -142,6 +142,19 @@ int transcode(struct transcode_thread *h,
...
   } else reopen_decoders = 0;
+
+  if (AV_HWDEVICE_TYPE_CUDA == ictx->hw_type && ictx->vi >= 0) {
+    if (ictx->last_format == AV_PIX_FMT_NONE) ictx->last_format = ictx->ic->streams[ictx->vi]->codecpar->format;
+    else if (ictx->ic->streams[ictx->vi]->codecpar->format != ictx->last_format) {
+      LPMS_WARN("Input pixel format has been changed in the middle.");
+      ictx->last_format = ictx->ic->streams[ictx->vi]->codecpar->format;
+      // free and reopen demuxer
+      free_input(&h->ictx);
+      ret = open_input(inp, &h->ictx);
```
yondonfu, line 153: For my own personal understanding:

Why does freeing & re-opening the demuxer eliminate the CUDA OOM error when the input pixel format changes in between segments? Does a change in the input pixel format trigger the error with the demuxer or the decoder? At a glance, it looks like `open_input()` opens the demuxer *and* the video/audio decoders so I'd like to understand which of the operations in `open_input()` are actually solving the issue here.

oscar-davids: [Here](https://github.com/livepeer/lpms/blob/e5409705d81e2ea3c009af81439eb4098e6fa128/ffmpeg/decoder.c#L244), AVCodecContext is initialized by the code parameter including Video Pixel Format.
If the demuxer is not reopened when the Video Pixel Format is changed, the demuxer tries HW decoding with the AVCodecContext initialized to a Pixel Format different from the Input Pixel Format.
This is exactly what causes the CUDA OOM error.

yondonfu: > If the demuxer is not reopened when the Video Pixel Format is changed, the demuxer tries HW decoding with the AVCodecContext initialized to a Pixel Format different from the Input Pixel Format.

Just to clarify - do you mean the decoder here? Seems like the problem is that the decoder needs to be re-opened via `open_video_decoder()` which is called by [open_input()](https://github.com/livepeer/lpms/blob/e5409705d81e2ea3c009af81439eb4098e6fa128/ffmpeg/decoder.c#L284). And `open_input()` does re-open the demuxer as well first, but IIUC the most important update is to re-open the decoder.

oscar-davids: Yes, Right. That mean decoder.

yondonfu: Got it. In that case, I recommend updating the comment here to make it clear that re-opening the decoder in `open_input()` is the important operation here to handle a change in the input pixel format.

darkdarkdragon: @oscar-davids 
Wouldn't calling `open_input` here result into second call to `open_audio_decoder` on line [164](https://github.com/livepeer/lpms/blob/a89fad9d802e922d3226dfa5e35e6e6360520e53/ffmpeg/transcoder.c#L164)? And that will lead to the memory leak?

oscar-davids: Fixed it as set reopen_decoders = 0;

darkdarkdragon (approved):
LGTM!

yondonfu (approved):
LGTM

PR #290, opened 2022-01-18
fast verification fix

darkdarkdragon:
Use latest FFmpeg commit
Add tests that cover latest fast verification fix


yondonfu (commented):
Looks good, but if we're deleting `install_ffmpeg.sh` can we update the README so that people know to download the `install_ffmpeg.sh` from go-livepeer?

PR #234, opened 2021-05-09
add transmuxer

darkdarkdragon:
just transmuxer, without transcoding

PR #292, opened 2022-01-18
add functions for video comparison

JamesWanglf:
### Step 1: C level comparison function[#234](https://github.com/livepeer/internal-project-tracking/issues/234)
- **For two segments, read stream data directly from the memory buffer at same time.**[#292](https://github.com/livepeer/internal-project-tracking/issues/292)
  We have defined two following functions in C level which are callable from outsite. 
  `int lpms_compare_video_bybuffer(void *buffer1, int len1, void *buffer2, int len2)`[here](https://github.com/livepeer/lpms/pull/292/files#diff-998114575433dc60658ff6542df0970e08cef7970396134aeedf868b90aace10R343)
  `int lpms_compare_video_bypath(char *vpath1, char *vpath2)`[here](https://github.com/livepeer/lpms/pull/292/files#diff-998114575433dc60658ff6542df0970e08cef7970396134aeedf868b90aace10R367)

  In the second function, the file paths of two different video segments will be used as input values, while the buffers and the buffer sizes will be used in the first function.

  These functions will return one integer. Only when the return value is equal to 1, it means two videos are same. Otherwise(0 or less than 0), it means those two are not same.
  Inside the second function, I read the file and save the file content into buffer by using the following function. After that, the first function will be called.
  `static uint8_t * get_filebuffer(const char *filename, int* fileLength)`[here](https://github.com/livepeer/lpms/pull/292/files#diff-998114575433dc60658ff6542df0970e08cef7970396134aeedf868b90aace10R198)

- **While reading buffer each segment, get metadata and count the number of Avpackets, compare time stamp.**[#293](https://github.com/livepeer/internal-project-tracking/issues/293)
  Two video will be compared with their width, height, bit_rate, packet count, XOR sum of avpacket PTS and XOR sum of hash(md5) of audio data.

  So we defined a struct [`match_info`](https://github.com/livepeer/lpms/pull/292/files#diff-998114575433dc60658ff6542df0970e08cef7970396134aeedf868b90aace10R8) for video information and function [`get_matchinfo`](https://github.com/livepeer/lpms/pull/292/files#diff-998114575433dc60658ff6542df0970e08cef7970396134aeedf868b90aace10R248) that we can fetch those information from the video content buffer.

  We get the metadata and necessary informations(packet-count, XOR of timestamp, XOR of md5 sum for audio data ) for a video comparison through [`get_matchinfo`](https://github.com/livepeer/lpms/pull/292/files#diff-998114575433dc60658ff6542df0970e08cef7970396134aeedf868b90aace10R248) function. 

  The following is a simple code snippet.
```
    info->packetcount++;
    info->timetamp ^= packet->pts;
    if(packet->stream_index == audioid && packet->size > 0) {
      av_md5_sum((uint8_t*)md5tmp, packet->data, packet->size);
      for (int i=0; i<4; i++) info->audiosum[i] ^= md5tmp[i];
    }
```
- **Meanwhile, run comparison of audio buffers with memcmp function.**[#293](https://github.com/livepeer/internal-project-tracking/issues/293)
  By using the [`get_matchinfo`](https://github.com/livepeer/lpms/pull/292/files#diff-998114575433dc60658ff6542df0970e08cef7970396134aeedf868b90aace10R248) function, we can get two struct variables for metadata of two videos.
  We compare video information in [here](https://github.com/livepeer/lpms/pull/292/files#diff-998114575433dc60658ff6542df0970e08cef7970396134aeedf868b90aace10R354).

### Step 2: Go level comparison function[#235](https://github.com/livepeer/internal-project-tracking/issues/235)
- **Write GO wrap API for lpms C level function.**
  We have defined two following functions in G level for lpms C level functions.
  `func CompareVideoByPath(fname1 string, fname2 string) (bool, error)`[here](https://github.com/livepeer/lpms/pull/292/files#diff-4ab68a4de67768f1acb5be217d4ec78bab27b99381025beedf82c11a5fa9c7aeR178)
  `func CompareVideoByBuffer(data1 []byte, data2 []byte) (bool, error)`[here](https://github.com/livepeer/lpms/pull/292/files#diff-4ab68a4de67768f1acb5be217d4ec78bab27b99381025beedf82c11a5fa9c7aeR199)

- **Unit test and benchmarking**
  By using [this](https://github.com/livepeer/lpms/pull/292/files#diff-13c64c7b8c1fc08e8f08edcf4e1b46d95ddbd51b93134f2cf896497569d62e60R1511) unit test, we can simply check whether the video comparison is working correctly or not.
  This test will output 3 files: out-1.ts, out-2.ts, out-3.ts.
  out-1.ts is same with out-2.ts, but different from out-3.ts.
  When we input out-1.ts and out-2.ts, it will return `true`.
  When we input out-1.ts and out-3.ts, it will return `false`.
  When we input out-1.ts and out-4.ts, it will return `false` and error `Can not open file out-4.ts`.

### DIFF
--- a/ffmpeg/api_test.go
+++ b/ffmpeg/api_test.go
@@ -2,6 +2,7 @@ package ffmpeg
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"testing"
 	"time"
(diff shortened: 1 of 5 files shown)


cyberj0g (approved):
Thanks for details, LGTM! Curios how fast is this verification compared to video duration?

oscar-davids (approved):
LGTM!

JamesWanglf:
> Thanks for details, LGTM! Curios how fast is this verification compared to video duration?

I have tested with lpms/transcoder/test.ts. It spent about 60ms for this 7s video duration to be compared.

darkdarkdragon (approved):
LGTM

PR #299, opened 2022-02-02
fix lpms_get_codec_info: use strlen to get string's length

darkdarkdragon:
`sizeof` is probably not what was intended


cyberj0g:
LGTM. Is there any special reason to not use `fmin`?

cyberj0g (approved):

darkdarkdragon:
@cyberj0g 
>  Is there any special reason to not use fmin?

No special reason, but it is unneeded function call, with type conversion. Actually I didn't removed it at first, but CircleCI complained that linker can't find `fmin` symbol, so I've removed it (some caching problem? Though, I think it doesn't matter).

PR #303, opened 2022-02-04
Add pixelFormat to GetCodecInfo()

AlexKordic:


ffmpeg/ffmpeg_test.go
```diff
@@ -1739,7 +1741,8 @@ func TestTranscoder_GetCodecInfo(t *testing.T) {
 		t.Errorf("Unexpected error %v", err)
 	}
 	fname = path.Join(wd, "..", "data", "bunny.mp4")
-	res, acodec, vcodec, err = GetCodecInfo(fname)
+	res, acodec, vcodec, pixelFormat, err = GetCodecInfo(fname)
+	fmt.Printf("bunny.mp4 %t %s %s %d %v\n", res, acodec, vcodec, pixelFormat, err)
```
AlexKordic, line 1745: Printf() shall be removed later

ffmpeg/videoprofile.go
```diff
@@ -246,17 +244,23 @@ func ParseProfiles(injson []byte) ([]VideoProfile, error) {
 			} else {
 				gopFloat, err := strconv.ParseFloat(profile.GOP, 64)
 				if err != nil {
-					return parsedProfiles, fmt.Errorf("Cannot parse the GOP value in the transcoding options: %w", err)
+					return parsedProfiles, fmt.Errorf("cannot parse the GOP value in the transcoding options: %w", err)
 				}
-				if gopFloat <= 0.0 {
-					return parsedProfiles, fmt.Errorf("Invalid gop value %f. Please set it to a positive value", gopFloat)
+				// gopFloat is allowed to be 0
+				if gopFloat < 0.0 {
```
AlexKordic, line 250: Tests in `go-livepeer` repo are designed to test for `0` in `profile.GOP`. Fixing this condition to allow 0

ffmpeg/ffmpeg.go
```diff
@@ -107,41 +108,162 @@ type TranscodeResults struct {
...
+	PixelFormatYUV444P12LE int = C.AV_PIX_FMT_YUV444P12LE
+)
+
+type ColorDepthBits int
+
+const (
+	ColorDepth8Bit  ColorDepthBits = 8
+	ColorDepth10Bit ColorDepthBits = 10
+	ColorDepth12Bit ColorDepthBits = 12
+)
```
oscar-davids, line 150: Maybe 16-bit color depth should be needed.

ffmpeg/extras.c
```diff
@@ -131,36 +135,45 @@ int lpms_rtmp2hls(char *listen, char *outf, char *ts_tmpl, char* seg_time, char
 //          1 for video with 0-frame, that needs bypass
 //          <0 invalid stream(s) or internal error
 //
```
oscar-davids, line 137: Need to update function descriptions.

ffmpeg/ffmpeg.go
AlexKordic: Good point. Can't imagine the display supporting 16bit HDR :)

ffmpeg/ffmpeg.go
```diff
@@ -107,41 +108,162 @@ type TranscodeResults struct {
...
+
+type GetCodecStatus int
+
+const (
+	GetCodecInternalError  GetCodecStatus = -1
+	GetCodecOk             GetCodecStatus = 0
+	GetCodecNeedsBypass    GetCodecStatus = 1
+	GetCodecStreamsMissing GetCodecStatus = 2
+)
+
```
oscar-davids, line 225: Variable type GetCodecStatus seems like a function name.
How about this?

type CodecStatus int 

const (
	CodecStatusInternalError  CodecStatus = -1
	CodecStatusOk             CodecStatus = 0
	CodecStatusNeedsBypass    CodecStatus = 1
	CodecStatusMissing CodecStatus = 2
)

ffmpeg/ffmpeg.go
```diff
@@ -506,11 +506,14 @@ func (t *Transcoder) Transcode(input *TranscodeOptionsIn, ps []TranscodeOptions)
...
-		// TODO: Check following condition, is vcodec == "" ?
-		videoMissing := status == CodecStatusNeedsBypass || vcodec == ""
-		if videoMissing {
-			// Audio-only segment, fail fast right here as we cannot handle them nicely
+		status, _, _, _, _ := GetCodecInfo(input.Fname)
+		// DONE: Check following condition, is vcodec == "" ?
+		// videoMissing := status == CodecStatusNeedsBypass || vcodec == ""
+		// if videoMissing {
+		if status == CodecStatusNeedsBypass {
+			// > no longer valid comment > Audio-only segment, fail fast right here as we cannot handle them nicely
```
AlexKordic, line 514: Hi @darkdarkdragon @cyberj0g, I need help on this change. 
The comment initially made sense but now regression was detected in `go-livepeer/core/transcoder_test.go::TestAudioCopy()`.

Its not clear to me what exactly `NeedsBypass` signals in the input media file? Depending on the answer we should keep change of this condition or change how tests use transcoder for audio only files.

Here are the links to original logic:
- [C function returning NeedsBypass](https://github.com/livepeer/lpms/blob/master/ffmpeg/extras.c#L156)
- [go function returning bool](https://github.com/livepeer/lpms/blob/master/ffmpeg/ffmpeg.go#L125)
- [original condition](https://github.com/livepeer/lpms/blob/master/ffmpeg/ffmpeg.go#L388)

With my interpretation being: returning error in case of audio-only segment. [`ErrTranscoderVid = errors.New("TranscoderInvalidVideo")`](https://github.com/livepeer/lpms/blob/master/ffmpeg/ffmpeg.go#L33)

Should we follow the intent of the comment `Audio-only segment, fail fast right here as we cannot handle them nicely` ? Or keep original logic and update comment appropriately?

oscar-davids (approved):
LGTM! Good work @AlexKordic

ffmpeg/ffmpeg.go
cyberj0g: From quirks.md:

We bypass the usual transcoding process for audio-only segments and just return them back as-is. To do this, we can check if a valid video stream is present without any actual video frames, and if so we skip transcoding.
While [this check](https://github.com/livepeer/lpms/blob/fe330766146dba62f3e1fccd07a4b96fa1abcf4d/ffmpeg/extras.c#L110-L117) is still implemented in LPMS, and was originally used by Transcoders, now this function is directly used by the Broadcaster since https://github.com/livepeer/go-livepeer/pull/1933.

In other words, audio-only segment shouldn't make its way into transcoder. The corresponding test should just check for correct error message.

ffmpeg/ffmpeg.go
```diff
@@ -107,41 +108,163 @@ type TranscodeResults struct {
...
+func GetCodecInfo(fname string) (CodecStatus, string, string, PixelFormat, error) {
 	var acodec, vcodec string
 	cfname := C.CString(fname)
 	defer C.free(unsafe.Pointer(cfname))
 	acodec_c := C.CString(strings.Repeat("0", 255))
 	vcodec_c := C.CString(strings.Repeat("0", 255))
 	defer C.free(unsafe.Pointer(acodec_c))
 	defer C.free(unsafe.Pointer(vcodec_c))
-	bres := int(C.lpms_get_codec_info(cfname, vcodec_c, acodec_c))
+	var params_c C.codec_info
```
cyberj0g, line 235: Should we free `params_c` struct too? Also I think eventually we'd want to return the struct from this function, not just use it internally in C interop.

ffmpeg/ffmpeg.go
```diff
@@ -165,9 +288,13 @@ func HasZeroVideoFrameBytes(data []byte) (bool, error) {
...
```
cyberj0g, line 291: Should we free `params_c` struct too? Would be great to avoid duplicating struct initialization snippet here.

ffmpeg/videoprofile.go
```diff
@@ -93,24 +93,26 @@ type VideoProfile struct {
...
-	P360p25fps16x9 = VideoProfile{Name: "P360p25fps16x9", Bitrate: "1000k", Framerate: 25, AspectRatio: "16:9", Resolution: "640x360"}
-	P360p30fps4x3  = VideoProfile{Name: "P360p30fps4x3", Bitrate: "1000k", Framerate: 30, AspectRatio: "4:3", Resolution: "480x360"}
-	P240p30fps16x9 = VideoProfile{Name: "P240p30fps16x9", Bitrate: "600k", Framerate: 30, AspectRatio: "16:9", Resolution: "426x240"}
-	P240p25fps16x9 = VideoProfile{Name: "P240p25fps16x9", Bitrate: "600k", Framerate: 25, AspectRatio: "16:9", Resolution: "426x240"}
-	P240p30fps4x3  = VideoProfile{Name: "P240p30fps4x3", Bitrate: "600k", Framerate: 30, AspectRatio: "4:3", Resolution: "320x240"}
-	P144p30fps16x9 = VideoProfile{Name: "P144p30fps16x9", Bitrate: "400k", Framerate: 30, AspectRatio: "16:9", Resolution: "256x144"}
-	P144p25fps16x9 = VideoProfile{Name: "P144p25fps16x9", Bitrate: "400k", Framerate: 25, AspectRatio: "16:9", Resolution: "256x144"}
+	P720p60fps16x9 = VideoProfile{Name: "P720p60fps16x9", Bitrate: "6000k", Framerate: 60, AspectRatio: "16:9", Resolution: "1280x720", ChromaFormat: ChromaSubsampling420, ColorDepth: ColorDepth8Bit}
```
cyberj0g, line 102: Maybe not initialize these fields to default enum values?

cyberj0g (changes requested):

ffmpeg/videoprofile.go
AlexKordic: I am not sure to understand yet what is the consequence? What is wanted behavior?

ffmpeg/ffmpeg.go
AlexKordic: Good question for deleting `params_c` struct. So far i have found the answer is: no explicit free on struct object created by go code.

Using IDE i can see the definition as:
```
type _Ctype_struct_s_codec_info struct {
	video_codec	*_Ctype_char
	audio_codec	*_Ctype_char
	pixel_format	_Ctype_int
	_		[4]byte
}
```
in (one can say generated file) `./build/0b/0bcf360880d821f4528cb9ca62dac5d996f2f71206d9362686c18ac37cae4b1f-d`

In this case go-created `params_c` struct is passed to C code to be populated. There are many constraints on how C code is allowed to manipulate given parameter and i think we are safe to do this.

Please correct me if i am wrong, `params_c` should be collected by go and allocated strings with `C.CString()` we free manually.

[Relevant research reference](https://stackoverflow.com/questions/10866195/stack-vs-heap-allocation-of-structs-in-go-and-how-they-relate-to-garbage-collec)

ffmpeg/ffmpeg.go
AlexKordic: Fixed

ffmpeg/ffmpeg.go
cyberj0g: Interesting! So we don't need to free struct explicitly in this case, because it's not explicitly allocated, this makes sense.

ffmpeg/videoprofile.go
AlexKordic: fixed

ffmpeg/ffmpeg.go
```diff
@@ -107,41 +108,163 @@ type TranscodeResults struct {
...
+const (
+	ChromaSubsampling420 ChromaSubsampling = iota
+	ChromaSubsampling422
+	ChromaSubsampling444
+)
+
+func (pixelFormat PixelFormat) Properties() (ChromaSubsampling, ColorDepthBits, error) {
+	switch pixelFormat.RawValue {
+	case C.AV_PIX_FMT_YUV420P:
+		return ChromaSubsampling420, 0, nil
```
cyberj0g, line 164: Since value doesn't have direct meaning anymore, can we use named constants for ColorDepthBits here?

ffmpeg/ffmpeg.go
```diff
@@ -107,41 +108,163 @@ type TranscodeResults struct {
...
+	PixelFormatYUV422P12BE int = C.AV_PIX_FMT_YUV422P12BE
+	PixelFormatYUV422P12LE int = C.AV_PIX_FMT_YUV422P12LE
+	PixelFormatYUV444P12BE int = C.AV_PIX_FMT_YUV444P12BE
+	PixelFormatYUV444P12LE int = C.AV_PIX_FMT_YUV444P12LE
+)
+
+type ColorDepthBits int
+
+const (
+	ColorDepth8Bit  ColorDepthBits = 0
```
cyberj0g, line 147: Maybe add a comment detailing how these values are related to actual number of bits?

cyberj0g (approved):

PR #305, opened 2022-02-09
Support encoder parameter in transoding options JSON

cyberj0g:
Another small part of https://github.com/livepeer/go-livepeer/issues/2123

Should be merged before go-livepeer PR, because it has codec name to enum conversion function.


darkdarkdragon (approved):
LGTM

Issue #301, opened 2022-02-03
TestTranscoder_DetectionFreq failing in CI


yondonfu:
The `TestTranscoder_DetectionFreq` seems to be consistently failing in CI with the following error:

```
[swscaler @ 0x7fc0fd514040] deprecated pixel format used, make sure you did set range correctly
[swscaler @ 0x7fc0fd514040] deprecated pixel format used, make sure you did set range correctly
[livepeer_dnn @ 0x7fc25dafdfc0] could not create DNN module for requested backend
[livepeer @ 0x7fc25c07b880] failed to initialize DNN backend
ERROR: transcoder.c:510] Invalid argument : Invalid argument
--- FAIL: TestTranscoder_DetectionFreq (2.72s)
    api_test.go:1459: DetectorInitializationError
panic: runtime error: invalid memory address or nil pointer dereference
	panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x744bd3]

goroutine 140 [running]:
testing.tRunner.func1.1(0x79e1c0, 0xa53fb0)
	/usr/local/go/src/testing/testing.go:1072 +0x30d
testing.tRunner.func1(0xc000103680)
	/usr/local/go/src/testing/testing.go:1075 +0x41a
panic(0x79e1c0, 0xa53fb0)
	/usr/local/go/src/runtime/panic.go:969 +0x1b9
github.com/livepeer/lpms/ffmpeg.(*Transcoder).StopTranscoder(0x0)
	/go/src/github.com/livepeer/lpms/ffmpeg/ffmpeg.go:616 +0x33
panic(0x79e1c0, 0xa53fb0)
	/usr/local/go/src/runtime/panic.go:969 +0x1b9
github.com/livepeer/lpms/ffmpeg.(*Transcoder).Transcode(0x0, 0xc000433c68, 0xc000180240, 0x2, 0x2, 0x0, 0x0, 0x0)
	/go/src/github.com/livepeer/lpms/ffmpeg/ffmpeg.go:312 +0x51
github.com/livepeer/lpms/ffmpeg.detectionFreq(0xc000103680, 0x0, 0x7fd726, 0x2)
	/go/src/github.com/livepeer/lpms/ffmpeg/api_test.go:1482 +0x42e
[...]

PR #306, opened 2022-02-11
Fix the failure of TestTranscoder_DetectionFreq in CircleCI

JamesWanglf:
This is related with #301 
- Before we run `TestTranscoder_DetectionFreq` function, we need to download `tasmodel.pb` first.
- We need to set the library path for the Tensorflow linking in env variables.
- To build the ffmpeg with libtensorflow, `experimental` should be contained in BUILD_TAGS.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,8 +20,8 @@ jobs:
             rm install_ffmpeg.sh || true
             wget https://raw.githubusercontent.com/livepeer/go-livepeer/master/install_ffmpeg.sh
 
-      # - restore_cache:
-      #     key: ffmpeg-cache-{{ checksum "install_ffmpeg.sh" }}
+      - restore_cache:
+          key: ffmpeg-cache-{{ checksum "install_ffmpeg.sh" }}
 
       - run:
           name: "Tweak PATH"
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -45,7 +45,13 @@ jobs:
             - "/home/circleci/ffmpeg"
             - "/home/circleci/compiled"
           key: ffmpeg-cache-{{ checksum "install_ffmpeg.sh" }}
-
+          
+      - run:
+          name: "Download tasmodel.pb"
+          command: |
+            wget https://github.com/livepeer/livepeer-ml/releases/download/v0.3/tasmodel.pb
```
cyberj0g, line 52: Given `lpms` is not released separately, maybe not pin the version here and link to [latest](https://github.com/livepeer/livepeer-ml/releases/latest/download/tasmodel.pb) release, so we don't have to update CI job for every new version of the model?

.circleci/config.yml
```diff
@@ -45,7 +45,13 @@ jobs:
...
+            mv tasmodel.pb ./ffmpeg
```
cyberj0g, line 53: Maybe download right to destination dir e.g. `wget URL -P ffmpeg`

cyberj0g (changes requested):

JamesWanglf:
If I disable `restore_cache` step, all tests will be passed.
If I enable this step, test will be failed at Test_DetectionFreq method. I guess this is occurred because the cached ffmpeg will be used, rather than the latest checkout.

oscar-davids (approved):
LGTM!

cyberj0g:
LGTM!

cyberj0g (approved):

PR #314, opened 2022-03-28
Fixed vscode debugging issue

oscar-davids:
Sometimes, the remote debugger broken in vs code and prints error message. [here ](https://discord.com/channels/423160867534929930/599227305230860289/956207429966786630)is the output.

### DIFF
--- a/ffmpeg/extras.c
+++ b/ffmpeg/extras.c
@@ -234,7 +234,7 @@ static uint8_t * get_filebuffer(const char *filename, int* fileLength)
     *fileLength = get_filesize(filename);
     if(*fileLength > 0) {
         // Cast to float is necessary to avoid int division
-        paddedLength = ceil(*fileLength / (float)AV_INPUT_BUFFER_PADDING_SIZE)*AV_INPUT_BUFFER_PADDING_SIZE + AV_INPUT_BUFFER_PADDING_SIZE;
+        paddedLength = (int)((*fileLength / (float)AV_INPUT_BUFFER_PADDING_SIZE)+1)*AV_INPUT_BUFFER_PADDING_SIZE + AV_INPUT_BUFFER_PADDING_SIZE;
         buffer = (uint8_t*)av_calloc(paddedLength, sizeof(uint8_t));
         if (!buffer) {
             av_log(NULL, AV_LOG_ERROR, "Could not allocate memory for reading signature file\n");


AlexKordic:
@thomshutt The failing tests are also present [here](https://app.circleci.com/pipelines/github/livepeer/lpms/469/workflows/b17c86fd-76b9-462a-a42f-d0a75ba1ce62/jobs/918) in [this commit](https://github.com/livepeer/lpms/commit/7c772da19b3372716d945d7a6bfaa1c5c08faa72)

thomshutt:
@AlexKordic Should we open a ticket to fix them then?

AlexKordic:
> @AlexKordic Should we open a ticket to fix them then?

yes

AlexKordic:
i managed to reproduce error message in the terminal
`go test -c -o /home/alex/livepeer/lpms/transcoder/__debug_bin -gcflags=all="-N -l" . in shell compiles fine as expected`
Command to break build is:
```
~/livepeer/lpms/transcoder$ CGO_CFLAGS=" -g3 " go test -c -o /home/alex/livepeer/lpms/transcoder/__debug_bin -gcflags=all="-N -l" .
# github.com/livepeer/lpms/ffmpeg
/usr/bin/ld: $WORK/b081/_x006.o: undefined reference to symbol 'ceil@@GLIBC_2.2.5'
/usr/bin/ld: /lib/x86_64-linux-gnu/libm.so.6: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
```
For some reason `CGO_CFLAGS` set cause `CGO_LDFLAGS` to be different ... unknown

Found by experiment following command:

`CGO_CFLAGS=" -g3 " CGO_LDFLAGS=" -l m " go test -c -o ./__debug_bin -gcflags=all="-N -l" .`
This `^^^` command works as expected

So i have added those 2 flags to settings.json of vscode:
```
{
....
    "go.toolsEnvVars": {
        "CGO_LDFLAGS" : " -l m ",
        "CGO_CFLAGS" : " -g3 "
    }
}
```

Not sure why link to libm is required explicitly. This configuration works for me now.

@oscar-davids what do you think should we update development documentation only or we still need to remove the `ceil()` ?

cyberj0g:
I have same issue locally and have to comment out `ceil` function every time. It seem to be caused by a combination of cgo and system glibc versions, which is problematic to resolve. Appeared after upgrading to Ubuntu 21.10 for me. Maybe just find replacement for `ceil`?

AlexKordic:
> Maybe just find replacement for `ceil`?

But what about other math functions we want to use in future? Are there any other options to fix this?

@cyberj0g Can you use `CGO_CFLAGS=" -g3 " CGO_LDFLAGS=" -l m " go test .....` command locally?

cyberj0g:
> Can you use CGO_CFLAGS=" -g3 " CGO_LDFLAGS=" -l m " go test ..... command locally?

Yes, adding -lm explicitly works. The question is why setting CGO_FLAGS for debugging cause missing -lm after I upgraded only system glibc? All other dependencies seem to be the same, maybe I'm missing something.

AlexKordic:
Can we use go build flags to add `-l m` to our project code?

thomshutt:
I agree with Alex that we should probably try to solve this at a higher level rather than trying to change which functions we use. Would adding the directives in the Go code that calls this work? https://pkg.go.dev/cmd/cgo

image

PR #316, opened 2022-03-31
segmenter: Remove invalid netcat flag in test

thomshutt:
This was causing tests to fail when I ran locally (M1 Mac) due to the flag being used wrongly for that version of Netcat:

https://github.com/st3fan/osx-10.9/blob/master/netcat-20/netcat.c#L271-L276

It also seems to not be a valid flag for some Linux versions:  https://linux.die.net/man/1/nc


oscar-davids (approved):
LGTM!

PR #317, opened 2022-04-01
Fix clip tests

AlexKordic:
Removed failing checks.

Didn't found good way to replace temporary failing checks.

Used manual test to verify proper operation:
- got input file timestamps: [frames_input.csv](https://github.com/livepeer/lpms/files/8397619/frames_input.csv)
- got encoded file timestamps: [frames_output.csv](https://github.com/livepeer/lpms/files/8397620/frames_output.csv)
- Produced video track contains proper frames
- Verified frames visually

Hard to implement above checks in code without additional image transformations. @oscar-davids any methods available we can use to verify correct set of frames are in output?

# Flushing procedure fix

When using tensorflow and flushing output in `process_out()` function `getmetadatainf()` call causes error to propagate and `TestTranscoder_DetectionFreq` test fails.

Added [condition](https://github.com/livepeer/lpms/blob/d309a54bcd234a5c7369a1e720c2c0d2d957ecc6/ffmpeg/encoder.c#L532) to cover this case.

### DIFF
--- a/doc/trim_output_tracks.md
+++ b/doc/trim_output_tracks.md
@@ -1,0 +1,55 @@
+
+# Trimming stream
+
+Use `From` and `To` parameters of `TranscodeOptions` to trim encoded output.
+
+Both audio and video tracks are trimmed from start if `From` is specified and trimmed at the end if `To` is specified. Starting point of `From` and `To` is first frame timestamp of corresponding track allowing to clip audio and video independently.
+
+Additional check is in place to skip all audio frames preceeding first video frame.
+
+## Usage
+
+```go
+	in := &TranscodeOptionsIn{Fname: "./input.ts"}
+	out := []TranscodeOptions{{
+		Profile: P144p30fps16x9,
+		Oname:   "./output.mp4",
+		From: 1 * time.Second,
+		To:   3 * time.Second,
+	}}
+	res, err := Transcode3(in, out)
+```
+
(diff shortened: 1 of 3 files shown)


doc/trim_output_tracks.md
```diff
@@ -0,0 +1,55 @@
...
+```
+V V V V V
+AAAAAAAAAA
+
+output:
+V V V 
+AAAAAA
+```
+
+offsted input:
```
thomshutt, line 35: offset?

doc/trim_output_tracks.md
```diff
@@ -0,0 +1,55 @@
...
+```
+    V V V V V
+AAAAAAAAAA   
+
+output:
+V V V
+ AAA
```
thomshutt, line 52: How come we don't get the audio frame that lines up with the first video frame in the output?

thomshutt (commented):
Looks fine to me - if @oscar-davids and @cyberj0g don't have any thoughts on improving the stability of these bits of the tests then I'd rather get rid of them and be able to rely on the test suite again

doc/trim_output_tracks.md
AlexKordic: There are 3 cases: both audio and video start on same timestamp or audio can start first or video can start first. In each case starting timestamp is not required to be 0 (common in splicing streams)

cyberj0g (approved):

PR #323, opened 2022-05-05
C configuration build changes

MikeIndiaAlpha:
The problem here was to ensue that C memory would not leak. Old code relied on Go's defer statements, but new code was no longer limited to single function scope. At first I though I may do checks first, allocation second but the problem remained because some of the outputs may get configured correctly (and C memory allocated) and then subsequent one may fail, triggering memory leaks. It could be done in two passes, first checks for all the outputs, then allocations but then some allocations share conditional logic with checks, so a lot of extra code.

I finally decided on the current approach - configuration creation function _always_ returns outputs array, and destroy function should _always_ be called. Then destroy function is releasing only the memory that actually was allocated. I am not super happy with this solution and would welcome suggestions how to make it better.

In addition, I did some small changes to make code easier to read - for example C allocations are grouped up close to where they are actually used.

### DIFF
--- a/ffmpeg/ffmpeg.go
+++ b/ffmpeg/ffmpeg.go
@@ -427,7 +427,9 @@ func newAVOpts(opts map[string]string) *C.AVDictionary {
 }
 
 // return encoding specific options for the given accel
-func configEncoder(inOpts *TranscodeOptionsIn, outOpts TranscodeOptions, inDev, outDev string) (string, string, error) {
+func configEncoder(inOpts *TranscodeOptionsIn, outOpts TranscodeOptions) (string, string, error) {
+	inDev := inOpts.Device
+	outDev := outOpts.Device
 	encoder := FfEncoderLookup[outOpts.Accel][outOpts.Profile.Encoder]
 	switch inOpts.Accel {
 	case Software:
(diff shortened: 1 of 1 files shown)


ffmpeg/ffmpeg.go
```diff
@@ -486,6 +488,253 @@ func Transcode3(input *TranscodeOptionsIn, ps []TranscodeOptions) (*TranscodeRes
...
+// to make sure that we don't leak C memory. Old code used defer statement to
+// pair every C allocation with corresponding release, but then the results
+// lifetime was obviously limited to the function scope - not something desired
+// here. Alternative solution would be to create configuration in two passes,
+// one just for checking, another one for actual C memory allocation - in effect
+// making code more complex. Hence "always destroy" policy, and destroy function
+// is removing only these fields that actually _were_ initialized
+func createCOutputParams(input *TranscodeOptionsIn, ps []TranscodeOptions) ([]C.output_params, error) {
+    params := make([]C.output_params, len(ps))
+    for i, p := range ps {
```
AlexKordic, line 502: The alternative to `defer` is [SetFinalizer](https://pkg.go.dev/runtime#SetFinalizer) which is bad choice here.

What can be improvement is to return "finalizer" to the caller instead of relying on comments to remind us in the future. Go unused variable error will remind us to defer call to returned finalizer.

```suggestion
func createCOutputParams(input *TranscodeOptionsIn, ps []TranscodeOptions) ([]C.output_params, func(), error) {
    params := make([]C.output_params, len(ps))
    finalizer := func () { destroyCOutputParams(params) } // always return as second return value
    for i, p := range ps {
```

cyberj0g: Good idea @AlexKordic. Another option, and, possibly, more Golang way to implement this is with Context. As this logic ultimately runs in response to user request, there should be a Context associated with it, which could be passed to any routine willing to perform cleanup action:
```
func createCOutputParams(ctx Context, input *TranscodeOptionsIn, ps []TranscodeOptions) {
	defer func() {
		<-ctx.Done()
		destroyCOutputParams(params)
	}()
}
// somewhere up the call stack
ctx, cancel := context.WithCancel()
defer cancel();

// usage
createCOutputParams(ctx, ...)
```
We may not have uninterrupted Context 'flow' everywhere currently, so it could be created right in `configEncoder()` I think.

AlexKordic: One question about go nature: what is difference between `defer func() {<-ctx.Done()}()` and `go func() {<-ctx.Done()}()` ? My feeling is `defer` would block before `createCOutputParams` returns and i would expect `go` to not affect return, @cyberj0g is this correct?

MikeIndiaAlpha: Thanks for your suggestions, guys. I read a bit about the Context approach and I really like it. However, I feel this code is going to change a lot in the future (for example I feel that this Nvidia specific filters stuff doesn't belong in high-level configuration code, but rather somewhere deep in pipeline implementation specialised for Nvidia HW). And thus I'd rather go with something simple. So I changed to what Alex suggested - I like that the compiler will remind the user should finaliser get forgotten.

Since there were no complaints I assume that I can turn that into ordinary PR.

ffmpeg/ffmpeg.go
```diff
@@ -529,217 +763,26 @@ func (t *Transcoder) Transcode(input *TranscodeOptionsIn, ps []TranscodeOptions)
...
-			if param.video.opts != nil {
-				C.av_dict_free(&param.video.opts)
-			}
-		}(&params[i])
-	}
+
+    // Output configuration
+	params, finalizer, err := createCOutputParams(input, ps)
+    // This prevents C memory leaks
+	defer finalizer();
```
AlexKordic, line 770: There are tabs instead of spaces here. Also `;` at the end of line is not wanted.

You can run `gofmt -w ./ffmpeg/ffmpeg.go` manually to apply all those formatting rules go community enforces.

I got nice IDE to do this on every file save. Awesome tooling.

AlexKordic (approved):
LGTM!

ffmpeg/ffmpeg.go
cyberj0g: Or just `go fmt ./...` in the project dir. We have it as a PR check in go-livepeer.

PR #329, opened 2022-05-23
Delete CircleCI workflow (replaced by GH Actions)

thomshutt:


yondonfu (approved):
LGTM

I deleted the CircleCI webhook for this repo so PRs should stop triggering CircleCI runs (which would fail because of the lack of a config file in the repo).

PR #332, opened 2022-06-07
core/vidplayer: add useful debug info to http response headers

emranemran:
When viewing HLS manifests/segments, it is often useful to view
information on how those segments were generated. This commit adds
fields such as git-sha and node identifiers to the HTTP response
headers.


emranemran:
Looks like a test correctly failed - will fix the test once the initial review is done.

leszko:
@emranemran Could you please change reviewers to include 2 pers only?

vidplayer/player.go
```diff
@@ -29,6 +29,20 @@ var ErrRTMP = errors.New("ErrRTMP")
...
+var (
+	lpVersion string 
+	nodeType string
+	nodeEthAddr string 
+)
+
+type VidPlayerDebug struct {
+	DebugEnabled	bool
+	LivepeerVersion string
+	NodeType	string
```
emranemran, line 42: TODO: remove `NodeType` here as we'll always be looking at the broadcaster node for async HLS playback.

PR #336, opened 2022-06-17
remove metadata comparison in the video comparison

oscar-davids:
For #[352](https://github.com/livepeer/internal-project-tracking/issues/352)
Regarding [this ](https://github.com/livepeer/lpms/pull/279#discussion_r778763259)changes, there are differences in GPU/CPU in bitrate, timestamp, and packet count. so, it removed the comparison of these.


cyberj0g (approved):
LGTM!

thomshutt (approved):
Looks good! If this doesn't break any tests here / in go-livepeer then we should probably add some

Issue #337, opened 2022-06-17
There are some audio missed segments during transcoding


oscar-davids:

yondonfu:
For the same source segment, if a T returns a rendition with audio and another T returns a rendition without audio then it sounds like there might be a general transcoding pipeline problem as opposed to a MPEG-7 video signature filter specific problem. If this is the case, we may want to take that source segment, identify the T that returned the rendition with missing audio and create a reproduction case such that by sending in the same source segment to the same T we can repeatedly get back a rendition without audio. Then, we can investigate why the audio track is being dropped and see if the reproduction case can be fixed.

oscar-davids:
**- Summary**
In fast verification phase 2, we compare two segments with width, height, and audio checksum, but sometimes we get audio-less segments from Ts. This greatly reduces the accuracy of the FV.

**- Reproduce issue**
We can reproduce the audio missing issue in a local environment when sending random segments with the audio or without audio to B.
 
step 1. prepare segment with audio and audio less segment to send to B
 step 2. execute B and O+T with livepeer binary
 step 3. first send audio-less segment to B, and send segments with audio using curl command
 step 4. We can see the transcoded segment directly or through "http://127.0.0.1:8936/stream/test.m3u8", we can check that there is no audio in the transcoded segments.

**- Why it happening?**

In the transcoding pipeline, we reuse the context of demuxer and muxer to reduce latency. When the T receives the audio-less segment first, T creates a demuxer context with it in [here](https://github.com/livepeer/lpms/blob/622b50738904a1c7d75a3b9650f1cf1341980670/ffmpeg/decoder.c#L307), but this context has no audio CodecContext. 
And then T receives another segment with audio, this time T gets audio information using pre-created context [here](https://github.com/livepeer/lpms/blob/622b50738904a1c7d75a3b9650f1cf1341980670/ffmpeg/decoder.c#L182), but this context has no audio CodecContext, so it returns incorrect information. T recognizes this segment for audio-less segment and returns audio-less renditions as result.
On other hand, when T receives segment with audio first and then receives audio-less segment , this segment has no audio packets so there is no problem in the transcoding pipeline.

**- Implementation**

we need to keep the audio codec information for the first received segment in T.  so need to add a member variable to `Transcoder` such as "lastacodec" [here]((https://github.com/livepeer/lpms/blob/622b50738904a1c7d75a3b9650f1cf1341980670/ffmpeg/ffmpeg.go#L87) ).
[...]

PR #341, opened 2022-06-29
reopen context when it added audio track into the segment

oscar-davids:
For fixing #337 

- added lastacodec variable in Transcoder structure
- call ["open_input"](https://github.com/livepeer/lpms/blob/622b50738904a1c7d75a3b9650f1cf1341980670/ffmpeg/decoder.c#L298) to recognize the audio track
- added unit test [here](https://github.com/livepeer/lpms/blob/e7ca14c09ef1e1302963b844376acd0dd8b1bc00/ffmpeg/api_test.go#L1575)

### DIFF
--- a/ffmpeg/transcoder.c
+++ b/ffmpeg/transcoder.c
@@ -439,6 +439,12 @@ int lpms_transcode(input_params *inp, output_params *params,
     if (ret < 0) {
       return ret;
     }
+  } else if (inp->reopendemux) {
+      free_input(&h->ictx);
+      ret = open_input(inp, &h->ictx);
+      if (ret < 0) {
+        return ret;
+      }
   }
 
   if (h->nb_outputs != nb_outputs) {
--- a/ffmpeg/api_test.go
+++ b/ffmpeg/api_test.go
@@ -3,6 +3,7 @@ package ffmpeg
 import (
 	"fmt"
 	"io/ioutil"
+	"math/rand"
 	"os"
 	"testing"
 	"time"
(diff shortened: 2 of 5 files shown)


ffmpeg/transcoder.c
```diff
@@ -439,6 +439,12 @@ int lpms_transcode(input_params *inp, output_params *params,
     if (ret < 0) {
       return ret;
     }
+  } else if (inp->reopendemux) {
+      free_input(&h->ictx);
+      ret = open_input(inp, &h->ictx);
```
cyberj0g, line 444: This will re-initialize NVDEC and add delay - could we just free audio context, and call `open_audio_decoder()`? I imagine that streams with some video-only segments are pretty common.

cyberj0g (changes requested):

ffmpeg/transcoder.c
cyberj0g: * So it's unlikely to be called mid-stream, but it still will be called if first segment doesn't have audio, and next segment has audio, is that right?
* Right, it's probably not so easy to inject audio demuxer dynamically, if possible at all. I wonder how would we address such issue withing low latency streaming workflow?

cyberj0g (approved):

ffmpeg/transcoder.c
oscar-davids: > So it's unlikely to be called mid-stream, but it still will be called if first segment doesn't have audio, and next segment has audio, is that right?

Right.

> Right, it's probably not so easy to inject audio demuxer dynamically, if possible at all. I wonder how would we address such issue withing low latency streaming workflow?

Yeah, I tried to inject audio demuxer dynamically, but couldn't find a reasonable path without reopening the segment.
open_audio_decoder() returns "No audio stream found in input" message.

MikeIndiaAlpha:
Guys, sorry for being late with my review, here it is anyway:
I don't like adding special parameter and code changes in transcoder.c. This is not the way to do it. We already have `lpms_transcoding_stop()` call which closes the demuxer/decoders. It is enough to call that function from Go code and desired effect will be achieved without any additional parameters/ifs. If the side effect of closing the outputs as well is not desired, I'd rather add `lpms_transcoding_close_input()` or similar to manipulate the Trancoder to that effect.

Please refrain from adding code to trancode.c this way. This is area I work on right now and this change conflicts with what I am doing now.
@oscar-davids @thomshutt - I'll be happy to add commit to master to make this change more in line with what I am doing with transcode.c if you guys don't mind.

PR #344, opened 2022-07-04
Demux reopen not in lpms_transcode()

MikeIndiaAlpha:
Due to my omission the previous solution was going against the
grain of upcoming refactoring changes in lpms_transcode().
Basically the idea is to have streamlined transcoder init() code
so that changes such as Low Latency can be implemented easily in
one place, instead of being distributed in many flows.

Besides, adding yet another "mode" to lpms_transcode() is not
really needed. The transcoder is kinda like any other object
instance - it retains its status between the lpms_transcode()
calls. So I think it is easier to add extra operations (such
as lpms_transcode_close_demux() introduced here) to change said
state, instead of increase the complexity of lpms_transcode()
call.

And finally, perhaps most important thing:
Changes like these are needed because LPMS doesn't really chave
good handling of changing configuration (in this case we are
talking about changing from container stream without audio to
the one with audio). It kind of pretends of doing so, and will
handle certain small differences kinda ok, but it is very easy
to come up with situation that will break it completely.

Ideally, I'd like to change that by reducing transient state
to the minimum (such as certain number of hardware buffers for
decoding and encoding) and really re-initialize everything else
that can be reinitialized cheaply. This however requires low
level hardware codec programming, it is not possible to do that
from ffmpeg level.

### DIFF
--- a/ffmpeg/ffmpeg.go
+++ b/ffmpeg/ffmpeg.go
@@ -852,8 +852,8 @@ func (t *Transcoder) Transcode(input *TranscodeOptionsIn, ps []TranscodeOptions)
 	if input == nil {
 		return nil, ErrTranscoderInp
 	}
-	var reopendemux C.int
-	reopendemux = 0
+	var reopendemux bool
+	reopendemux = false
 	// don't read metadata for pipe input, because it can't seek back and av_find_input_format in the decoder will fail
 	if !strings.HasPrefix(strings.ToLower(input.Fname), "pipe:") {
 		status, format, err := GetCodecInfo(input.Fname)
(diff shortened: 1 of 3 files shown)


ffmpeg/ffmpeg.go
```diff
@@ -943,6 +949,16 @@ func (t *Transcoder) Transcode(input *TranscodeOptionsIn, ps []TranscodeOptions)
 		paramsPointer = (*C.output_params)(&params[0])
 		resultsPointer = (*C.output_results)(&results[0])
 	}
+	if reopendemux {
+		// forcefully close and open demuxer
+		ret := int(C.lpms_transcode_reopen_demux(inp))
+    if ret != 0 {
```
oscar-davids, line 955: nitpick: align space before merge here.

MikeIndiaAlpha: My bad, fixed.

PR #345, opened 2022-07-09
Experimental queued implementation for demuxer input

MikeIndiaAlpha:
This is an experimental - but working, and all the tests are passing - for queued demuxer input. The idea is to use it with other Low Latency changes, to support network input to transcoding pipeline. If the arrangements developed here will be found satisfactory, output counterpart should be created (for muxers).

I am not sure of the following:
- Go code - not much of it here but still perhaps style could be better
- Seek function implementation - hack seems to work but still

I am pretty satisfied with overall approach but still feel that it can be improved.


AlexKordic:
I vote for changing `lpms_add_input_buffer()` to block until given `StreamBuffer *input` is consumed. This would allow to establish proper pace. For example go code would read from disk or network and give bytes/buffer to transcoder, repeat until EOF. Without blocking input call go code doesn't know what read pace to employ. Alternative synchronization designs seam too complicated to me.

Please also consider simplifying input structure. Can we remove `StreamBufferQueue` in favour of single `StreamBuffer` in flight? The partial frame can be cached within transcode component, to hide it from go code.

AlexKordic:
@MikeIndiaAlpha question regarding seek: For what do we need seeking for? VOD?

AlexKordic:
Please include timestamp in the output `StreamBuffer` structure.

MikeIndiaAlpha:
> I vote for changing `lpms_add_input_buffer()` to block until given `StreamBuffer *input` is consumed. This would allow to establish proper pace. For example go code would read from disk or network and give bytes/buffer to transcoder, repeat until EOF. Without blocking input call go code doesn't know what read pace to employ.

I can't agree in case of "read from network" scenario. Please consider this situation: we are receiving data from network as a series of buffers, each of them holding (say) one-tenth of a frame. I believe this is realistic given that for network transfer environmental stuff like MTU can have nearly arbitrary effects. Now, let us consider situation in which decoder is at work at previous frame. Network layer gives us 1/10th of a frame, we send it to the decoder and that call blocks (as you suggested).

The block will take until current frame will get decoded, then (and only then) decoder "sees" the input buffer, picks it up, tries to decode but alas - this is just 1/10th of a frame. So it cannot work, it blocks on the input...and so it continues, with blocking of decoder thread 10 times before it can even start decoding. I don't believe this helps the latency, and this change is all about the low latency, right?

Now consider proper "queued" situation - the network layer will be able to accumulate all the buffers as they arrive, and the decoder is likely to see complete data for the next frame the first time it tries to decode. So no time wasted waiting for input, it will just jump right into decoding next frame, and we are close to saturating the decoder which is a key to efficiency here (because without decoded frame all the remaining parts of the pipeline are just idling).

Moveover, some HW decoder really don't like being given parts of the frame with codecs like h.264. Some of them will always start to decode, and if frame is not complete they'll just reset. So if you give them frame in, like, 10 pieces, your are going to make your decoder do some 5x the work or so.

TBH I just think that the I/O here will be "naturally" timed by the network. I do see your point in the "file" case, however. We might have a big files, and we may not want to load them completely. But to solve this issue...

> Alternative synchronization designs seam too complicated to me.

...nah, I could easily add "lpms_add_input_buffer_and_block" call for use with the files. All I need is another condition variable, perhaps five lines of code total. We could then use nonblocking version for network I/O, and blocking one for files. What do you think?
 
> Please also consider simplifying input structure. Can we remove `StreamBufferQueue` in favour of single `StreamBuffer` in flight? The partial frame can be cached within transcode component, to hide it from go code.

Like I wrote, I consider queue essential to low latency. Also, on the output side we have multiple muxers. The idea is to have all the packets from different muxers in the same output queue, and index field will be used to separate the packets.

MikeIndiaAlpha:
> @MikeIndiaAlpha question regarding seek: For what do we need seeking for? VOD?

We need seeking for demuxing. When I plugged my I/O I had seek returning -1, I also tried disabling seeking by setting the `seeable` field of AVIOContext. Unfortunately a lot of our tests simply failed to run then. It seems that depending on the container format (and so on the demuxer) seek() call may be needed. And it seems to be used in different ways for different formats, for example some will just seek forward with SEEK_SET, and some will seek backward with SEEK_END. I had to add those for the tests to pass.

Also I remember that the go code used some probing function that had to be disabled with pipes, because you cannot seek with a pipe. I think if we moved data from the pipe to queue then I could expose similar probing function working on queue's content and it would work just fine thanks to seek capabilities. Perhaps it could be useful?

PR #350, opened 2022-07-16
change video comparison function

oscar-davids:
- Hardware transcoding fails when livepeer is sent segments with very few video frames. It works fine when running software trascoding but fails in hardware trascoding. This is caused because internal buffers of Nvidia's decoder are bigger compared to software decoder, current to solve this issue we defined [SENTINEL_MAX ](https://github.com/livepeer/lpms/blob/fb9f13d91fc2b20e51b799216c322dbfc30f07c5/ffmpeg/decoder.h#L33)= 5, but still appears issue on other GPU(i.e. 1660).
So changed  SENTINEL_MAX to 8.

- Rarely, the first audio packet is dropped during transcoding. Therefore, do not compare the first audio packets those timestamp equals 0 [here](https://github.com/livepeer/lpms/blob/fb9f13d91fc2b20e51b799216c322dbfc30f07c5/ffmpeg/extras.c#L344).

### DIFF
--- a/ffmpeg/decoder.h
+++ b/ffmpeg/decoder.h
@@ -30,7 +30,7 @@ struct input_ctx {
   // We maintain a count of sentinel packets sent without receiving any
   // valid frames back, and stop flushing if it crosses SENTINEL_MAX.
   // FIXME This is needed due to issue #155 - input/output frame mismatch.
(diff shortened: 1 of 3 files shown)


MikeIndiaAlpha (approved):
Looks good to me!

PR #357, opened 2022-11-17
Ensure signature comparison success ratio among Nvidia and CPU signatures

cyberj0g:
https://github.com/livepeer/go-livepeer/issues/2654

**Changes**
* Add signature files calculated during Nvidia and CPU transcoding
* Add a unit test, which calculates and ensures that false negative (signature mismatch for matching renditions) ratio is less than 1% between Nvidia and CPU signatures
* Add a unit test, which calculates and ensures that false positive (signature match for different renditions) ratio is less than 15% for both CPU and Nvidia signatures

For this unit test to pass, a [PR](https://github.com/livepeer/FFmpeg/pull/30/) which sets threshold to detect frames as similar in signature comparison function of Ffmpeg needs to be merged first.

Below are results for the threshold value of 116 (default) and 150 (new value). I think, such an increase in FP ratio is acceptable. New value was picked by hand in a few attempts, to reduce the chance of overfitting our limited test dataset.
```
th_xh (thl1) = 116
----------------
CPU 360p False Positive rate: 0.091133
CPU 720p False Positive rate: 0.110837
Nvidia 360p False Positive rate: 0.091133
Nvidia 720p False Positive rate: 0.110837
CPU - Nvidia 360p False Negative rate: 0.068966
CPU - Nvidia 720p False Negative rate: 0.000000

th_xh (thl1) = 150
----------------
CPU 360p False Positive rate: 0.103448
CPU 720p False Positive rate: 0.125616
Nvidia 360p False Positive rate: 0.113300
Nvidia 720p False Positive rate: 0.128079
CPU - Nvidia 360p False Negative rate: 0.000000
CPU - Nvidia 720p False Negative rate: 0.000000
```


ffmpeg/sign_test.go
```diff
@@ -169,3 +171,109 @@ func Test_SignDataCompare(t *testing.T) {
...
+
+	fmt.Printf("----------------\n")
+	fmt.Printf("CPU 360p False Positive rate: %f\n", cpu360FpRatio)
+	fmt.Printf("CPU 720p False Positive rate: %f\n", cpu720FpRatio)
+	fmt.Printf("Nvidia 360p False Positive rate: %f\n", nv360FpRatio)
+	fmt.Printf("Nvidia 720p False Positive rate: %f\n", nv720FpRatio)
+	fmt.Printf("CPU - Nvidia 360p False Negative rate: %f\n", cpuNv360FnRatio)
+	fmt.Printf("CPU - Nvidia 720p False Negative rate: %f\n", cpuNv720FnRatio)
+
+	assert.True(t, cpuNv720FnRatio <= 0.01)
```
yondonfu, line 272: Could cleanup assertions a bit with:

```
assert := assert.New(t)
assert.True(...)
assert.True(...)
```

ffmpeg/sign_test.go
```diff
@@ -169,3 +171,109 @@ func Test_SignDataCompare(t *testing.T) {
...
+	assert.True(t, cpuNv360FnRatio <= 0.01)
```
yondonfu, line 273: Can we define 0.01 and 0.15 as constants somewhere so it is clear that they are target FNR and FPR?

yondonfu (changes requested):
Curious if generating signatures instead of checking them in would be possible since that would allow us to do comparisons of signatures generated using the current ffmpeg impl being used? Thoughts?

cyberj0g:
Thanks @yondonfu, all comments are addressed. We can generate signatures right in the test, since `lpms` tests are running on GPU runner now.

yondonfu (approved):
LGTM!

PR #371, opened 2023-08-29
Improve Transcode Quality

leszko:
Improve Transcode Quality with:
- Always setting `-preset slow -tier high`
- Allowing setting CRF (CQ) to use constant rate factor instead of constant bitrate

Note that there will be a separate PR to change `scale_cuda` to `scale_npp`.

Related:
- [Design Doc](https://www.notion.so/livepeer/Transcode-Quality-investigation-improvements-11314327aa484fb6927a48b102bd54bf#b821f216259b40d78ebbff164d67c062)
- [Linear Issue](https://linear.app/livepeer/issue/VID-380/lpms-apply-proposed-ffmpeg-parameter-changes)


thomshutt (approved):

leszko:
CI is failing, but it's unrelated to this PR, failing also on master: https://github.com/livepeer/lpms/actions/runs/6022454830/job/16337223769?pr=372

ffmpeg/ffmpeg.go
```diff
@@ -694,6 +694,24 @@ func createCOutputParams(input *TranscodeOptionsIn, ps []TranscodeOptions) ([]C.
...
+			if p.Profile.CRF != 0 {
+				if p.Profile.CRF <= 63 {
+					p.VideoEncoder.Opts["crf"] = strconv.Itoa(int(p.Profile.CRF))
+				} else {
+					glog.Warning("Cannot use CRF param, value out of range (0-63)")
+				}
+
+				// There's no direct numerical correspondence between CQ and CRF.
+				// From some experiments, it seems that setting CQ = CRF + 7 gives similar visual effects.
+				cq := p.Profile.CRF + 7
```
emranemran, line 709: How did you end up with this formula? Lower CQ values typically indicate better quality at expense of compressions. Also encoders typically will pick either CQ and CRF as the main control so I'm not sure if CQ here will actually make a difference.

leszko: CRF is used by CPU-based transcoding, CQ is used by GPU-based transcoding.
So, only one of these values will be used at the time.

Technically there is no direct numerical correspondence between CQ and CRF, so the other option would be to set both CRF and CQ in our API. But I think it's not a pragmatic approach right now. Most of our transcoders are GPU anyway.

How did I come with `cq  := p.Profile.CRF + 7`, just experimenting with videos. But obviously not for all videos it will give the same results, since technically CRF and CQ values do not correspond to each other.

emranemran: The `cq  := p.Profile.CRF + 7` doesn't quite make sense to me esp since they don't correspond to one another. Also if CRF is CPU only, then why do we need it?

leszko: Explained here: https://discord.com/channels/423160867534929930/1140578841136484402/1149041283583721634

lets discuss more in discord

PR #374, opened 2023-09-08
Rename profile.CRF to profile.Quality

leszko:
Since we set CQ based on the `quality`, then to avoid confusion, let's not use the `crf` name.

Discussed more in [Discord](https://discord.com/channels/423160867534929930/1140578841136484402/1149583985530380380).


thomshutt (approved):
LGTM - tests are failing but doesn't seem related

PR #377, opened 2024-01-12
Remove scene detection code

thomshutt:
This is no longer needed and so is being removed to make the code more managable


leszko (approved):
Good work @thomshutt . Looks great to remove soo much unused comment.

One question, do we maybe want to check if it still works fine on an Nvidia machine?

thomshutt:
@leszko I've been running a version of go-livepeer with this LPMS commit and it seems fine

PR #378, opened 2024-01-18
workflows: Try to get lpms builds in line with go-livepeer builds

hjpotter92:


thomshutt:
@hjpotter92 I think the issue is that it's trying to run Nvidia tests, but we don't have the hardware / libraries on these boxes: ```[AVHWDeviceContext @ 0x7f60500051c0] Cannot load libcuda.so.1
[AVHWDeviceContext @ 0x7f60500051c0] Could not dynamically load CUDA```

thomshutt (approved):

PR #393, opened 2024-04-11
Improve VFR support

j0sh:
#### Improve VFR support ( 55a5d805d32927d3e6433a64fa1ad343b1370e61 )
Manually calculate the duration of each frame and set
the PTS to that before submitting to the filtergraph.

This allows us to better support variable frame rates,
and is also better aligned with how ffmpeg does it.

This may change the number of frames output by the FPS
filter by +/- 1 frame. These aren't issues in themselves
but breaks a lot of test cases which will need to be updated.

#### Notes

Previously we were depending on a calculated frame rate ([r_frame_rate](https://github.com/FFmpeg/FFmpeg/blob/dd5f665b4010f8a0142ce3cba3305b173eb37dfe/libavformat/avformat.h#L903-L911))
but this isn't set to a sensible value for VFR streams. Packet duration also
can't be relied on (see the new VFR test case where each duration is N/A),
so we need to do some manual bookkeeping to calculate the timestamp
difference between frames ourselves.

I haven't yet fully understood what happens with the FPS filter during the
transition between segments but the behavior seems okay and the
timestamps don't change too much, aside from the +/- 1 frame difference.

Most unit tests pass on my machine, except I did not test GPU or non-H264
codecs. I don't *think* things would be much different there, but someone with
a GPU should check. The tests in `ffmpeg/sign_test.go` all timed out for me
and there is a segmenter test that seems flaky, but these are probably unrelated.

### DIFF
--- a/ffmpeg/api_test.go
+++ b/ffmpeg/api_test.go
@@ -51,7 +51,7 @@ func TestAPI_SkippedSegment(t *testing.T) {
 		if res.Decoded.Frames != 120 {
 			t.Error("Did not get decoded frames", res.Decoded.Frames)
 		}
-		if res.Encoded[1].Frames != 246 {
+		if res.Encoded[1].Frames != 245 {
 			t.Error("Did not get encoded frames ", res.Encoded[1].Frames)
 		}
 	}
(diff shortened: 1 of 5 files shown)


j0sh:
Another quick note that I forgot to mention. I am not sure how this fix would be rolled-out operationally, but because this will produce different frame counts (and thus different timings), it is important that streams be "pinned" to the same version of LPMS if some transcoders are slow to update themselves. Otherwise there might be some breakage in between segment intervals, eg timestamp overlaps or gaps.

leszko (approved):

PR #400, opened 2024-05-13
Add an error return to makeStreamID to make issues explicit

thomshutt:
This is to allow us to signal different types of errors in go-livepeer and ultimately pass back different HTTP statuses from the Broadcaster depending on the stream creation failure reason

### DIFF
--- a/vidlistener/listener.go
+++ b/vidlistener/listener.go
@@ -36,7 +36,7 @@ func (self *VidListener) HandleRTMPPublish(
 			glog.V(2).Infof("RTMP server got upstream: %v", conn.URL)
 
 			strmID, err := makeStreamID(conn.URL)
-			if err != nil || strmID.StreamID() == "" {
+			if err != nil || strmID == nil || strmID.StreamID() == "" {
 				conn.Close()
 				return
 			}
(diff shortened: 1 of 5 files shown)


vidlistener/listener.go
```diff
@@ -22,21 +22,21 @@ type VidListener struct {
...
+	makeStreamID func(url *url.URL) (strmID stream.AppData, err error),
 	gotStream func(url *url.URL, rtmpStrm stream.RTMPVideoStream) error,
 	endStream func(url *url.URL, rtmpStrm stream.RTMPVideoStream) error) {
 
 	if self.RtmpServer != nil {
 		self.RtmpServer.HandlePublish = func(conn *joy4rtmp.Conn) {
 			glog.V(2).Infof("RTMP server got upstream: %v", conn.URL)
 
-			strmID := makeStreamID(conn.URL)
-			if strmID == nil || strmID.StreamID() == "" {
```
leszko, line 39: Where is this `makeStreamID()` actually implemented? In go-livepeer? I wonder if changing this condition here won't bring any regression.

thomshutt: Yep, in `go-livepeer`. I think that we can probably keep the original check as well though if we'd like to be safe

leszko (approved):

mjh1 (approved):

PR #405, opened 2024-07-01
Fix FPS passthrough

j0sh:
### Do not merge

This PR depends on newer ffmpeg API features (copying user-supplied opaque data from frame to pkt) so the changes will have to be merged alongside the ffmpeg upgrade itself - #406. Posting a PR here so the details of the fix are more easily accessible as a historical reference

---

* Set the encoder timebase using AVCodecContext.framerate instead of the decoder's AVCodecContext.time_base.

  The use of AVCodecContext.time_base is deprecated for decoding. See https://ffmpeg.org/doxygen/3.3/structAVCodecContext.html#ab7bfeb9fa5840aac090e2b0bd0ef7589

* Adjust the packet timebase as necessary for FPS pass through to match the encoder's expected timebase. For filtergraphs using FPS adjustment, the filtergraph output timebase will match the framerate (1 / framerate) and the encoder is configured for the same.

  However, for FPS pass through, the filtergraph's output timebase will match the input timebase (since there is no FPS adjustment) while the encoder uses the timebase detected from the decoder's framerate. Since the input timebase does not typically match the FPS (eg 90khz for mpegts vs 30fps), we need to adjust the packet timestamps (in container timebase) to the encoder's expected timebase.

* For the specific case of FPS passthrough, preserve the original PTS as much as possible since we are trying to re-encode existing frames one-to-one. Use the opaque field for this, since it is already being populated with the original PTS to detect sentinel packets during flushing.

  Without this, timestamps can be slightly "squashed" down when rescaling output packets to the muxer's timebase, due to the loss of precision (eg, demuxer 90khz -> encoder 30hz -> muxer 90khz)

### DIFF
--- a/ffmpeg/encoder.c
+++ b/ffmpeg/encoder.c
@@ -241,8 +241,9 @@ int open_output(struct output_ctx *octx, struct input_ctx *ictx)
     else if (ictx->vc->framerate.num && ictx->vc->framerate.den) vc->framerate = ictx->vc->framerate;
     else vc->framerate = ictx->ic->streams[ictx->vi]->r_frame_rate;
     if (octx->fps.den) vc->time_base = av_buffersink_get_time_base(octx->vf.sink_ctx);
-    else if (ictx->vc->time_base.num && ictx->vc->time_base.den) vc->time_base = ictx->vc->time_base;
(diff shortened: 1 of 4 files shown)


j0sh:
Incorporated in #406

PR #407, opened 2024-07-05
Add media duration to lpms_get_codec_info for GetCodecInfo

eliteprox:
Adds the duration to `lpms_get_codec_info` for audio only and audio/video files.

This is needed for the audio to text pipeline


cmd/test/get_fps_dur.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"fmt"
+	"os"
+	"text/tabwriter"
+	"time"
+
+	"github.com/livepeer/lpms/ffmpeg"
+)
+
+func main() {
```
leszko, line 12: Can you change to a proper unit test?

ffmpeg/extras.c
```diff
@@ -163,6 +188,13 @@ int lpms_get_codec_info(char *fname, pcodec_info out)
   if(!audio_present && !video_present) {
     // instead of returning -1
     ret = GET_CODEC_STREAMS_MISSING;
+  } else if (video_present) {
+    out->dur = ic->duration / AV_TIME_BASE;
+  } else if (audio_present && !video_present) {
+    out->dur = calculate_duration(ic, astream);
```
leszko, line 194: Why do we calculate duration only if video is not present?

leszko (commented):
Thanks for the PR. Added 2 comments.

Other than that, @j0sh could you also look at the PR (since I believe you have the best knowledge about the lpms project)?

ffmpeg/extras.c
eliteprox: During development I found that `out->dur = ic->duration / AV_TIME_BASE;` was returning a very large negative number that wasn't useful. The `calculate_duration` function is a more general, but less efficient means of calculating the duration. It seemed logical to keep this only to audio-only. The function name is a bit ambiguous, should I rename it?

leszko: Yeah, better to rename it. Plus, maybe add this comment into the code.

eliteprox: Renamed `calculate_duration` to `calculate_audio_dur` and added the comment

cmd/test/get_fps_dur.go
eliteprox: Moved to [ffmpeg/ffmpeg_test.go](https://github.com/livepeer/lpms/pull/407/files#diff-4ab68a4de67768f1acb5be217d4ec78bab27b99381025beedf82c11a5fa9c7ae)

ffmpeg/ffmpeg_test.go
```diff
@@ -1869,3 +1870,25 @@ func TestResolution_Clamp(t *testing.T) {
 	checkError = require.Error
 	test(l, Size{300, 600}, portrait, Size{300, 600})
 }
+
+func TestGetFPSDuration(t *testing.T) {
```
j0sh, line 1874: this does not seem to actually test anything? please see how other testcases are constructed

ffmpeg/ffmpeg.go
```diff
@@ -245,6 +245,8 @@ type MediaFormatInfo struct {
 	Acodec, Vcodec string
 	PixFormat      PixelFormat
 	Width, Height  int
+	FPS            float32
+	Dur            float32
```
j0sh, line 249: not sure that float32 is the best type for either of these, but if you don't care about precision it might be ok. i would opt for a rational for fps and int64 for (scaled) duration

ffmpeg/extras.c
```diff
@@ -149,6 +172,9 @@ int lpms_get_codec_info(char *fname, pcodec_info out)
 #define MIN(x, y) (((x) < (y)) ? (x) : (y))
   AVFormatContext *ic = NULL;
   AVCodec *ac, *vc;
+
+  int64_t last_pts = AV_NOPTS_VALUE;
```
j0sh, line 176: unused

ffmpeg/extras.c
```diff
@@ -130,6 +130,29 @@ int lpms_rtmp2hls(char *listen, char *outf, char *ts_tmpl, char* seg_time, char
   return ret == AVERROR_EOF ? 0 : ret;
 }
 
+//Calculates the duration of audio stream by counting the number of audio packets
+double calculate_audio_dur(AVFormatContext *ic, int astream) {
+  AVPacket pkt;
+  av_init_packet(&pkt);
+  double duration = 0;
+  int64_t last_pts = AV_NOPTS_VALUE;
+  while (av_read_frame(ic, &pkt) >= 0) {
```
j0sh, line 139: this will require reworking if another part of GetCodecInfo ever requires another pass through the file

ffmpeg/extras.c
```diff
@@ -186,6 +221,7 @@ int lpms_get_codec_info(char *fname, pcodec_info out)
       // Indicate failure to extract audio codec from given container
       out->audio_codec[0] = 0;
   }
+
```
j0sh, line 224: stray

ffmpeg/ffmpeg_test.go
eliteprox: Rewrote into two tests which validate FPS and duration. Added test files. https://github.com/livepeer/lpms/pull/407/files#diff-aeb5ed5cb59cc9ff3ded5fa0bebee480bb3273b66c016f159c80c188a6b21611

ffmpeg/extras.c
eliteprox: I added `av_seek_frame` to reset the file position. Will this work? https://github.com/livepeer/lpms/pull/407/files#diff-998114575433dc60658ff6542df0970e08cef7970396134aeedf868b90aace10R140

ffmpeg/ffmpeg.go
eliteprox: It's a good idea, I think float32 is sufficient precision for what we need right now

ffmpeg/extras.c
eliteprox: Removed un-used variable

ffmpeg/extras.c
eliteprox: Removed stray line break

ffmpeg/extras.c
```diff
@@ -176,6 +208,7 @@ int lpms_get_codec_info(char *fname, pcodec_info out)
       }
       out->width  = ic->streams[vstream]->codecpar->width;
       out->height = ic->streams[vstream]->codecpar->height;
+      out->fps = av_q2d(ic->streams[vstream]->avg_frame_rate);
```
j0sh, line 211: Note that [`avg_frame_rate`](https://ffmpeg.org/doxygen/trunk/structAVStream.html#a946e1e9b89eeeae4cab8a833b482c1ad)  is not guaranteed to be set to anything. If unset, [`r_frame_rate`](https://ffmpeg.org/doxygen/trunk/structAVStream.html#ad63fb11cc1415e278e09ddc676e8a1ad) might be acceptable but I am not sure how this value would be used.

j0sh (changes requested):

ffmpeg/ffmpeg_test.go
```diff
@@ -1869,3 +1869,65 @@ func TestResolution_Clamp(t *testing.T) {
...
+		}
+
+		_, format, err = GetCodecInfoBytes(data)
+		if err != nil {
+			t.Error(err)
+		}
+
+		took := time.Since(start).Milliseconds()
+		fmt.Printf("%v\ttook=%v\tdur=%v\tacodec=%v\n", fname, took, format.Dur, format.Acodec)
+		if format.Dur <= float32(0) {
```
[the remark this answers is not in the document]
eliteprox: Updated test to generate test files and verify expected duration and fps values

ffmpeg/extras.c
```diff
@@ -130,6 +130,31 @@ int lpms_rtmp2hls(char *listen, char *outf, char *ts_tmpl, char* seg_time, char
...
 }
 
+//Calculates the duration of audio stream by counting the number of audio packets
+double calculate_audio_dur(AVFormatContext *ic, int astream) {
+  AVPacket pkt;
+  av_init_packet(&pkt);
+  double duration = 0;
+  int64_t last_pts = AV_NOPTS_VALUE;
+  // Seek to the beginning of the audio stream
+  av_seek_frame(ic, astream, 0, AVSEEK_FLAG_BACKWARD);
```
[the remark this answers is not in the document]
eliteprox: I'm fine with removing this line, it will probably cause more trouble than good

ffmpeg/ffmpeg_test.go
```diff
@@ -1869,3 +1869,65 @@ func TestResolution_Clamp(t *testing.T) {
...
 	test(l, Size{300, 600}, portrait, Size{300, 600})
 }
+
+func TestDuration_GetCodecInfo(t *testing.T) {
+	wd, err := os.Getwd()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	files := []string{"bunny.mp4", "duplicate-audio-dts.ts", "bunny.mp3", "bunny.webm", "bunny.flac", "bunny.m4a"}
```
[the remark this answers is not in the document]
eliteprox: Removed the test files and updated to generate them

ffmpeg/ffmpeg_test.go
```diff
@@ -1869,3 +1869,45 @@ func TestResolution_Clamp(t *testing.T) {
...
+
+		_, format, err = GetCodecInfoBytes(data)
+		if err != nil {
+			t.Error(err)
+		}
+
+		took := time.Since(start).Milliseconds()
+		fmt.Printf("%v\ttook=%v\tdur=%v\fps=%v\tvcodec=%v\tacodec=%v\n", fname, took, format.DurSecs, format.FPS, format.Vcodec, format.Acodec)
+		assert.True(t, format.DurSecs == file.Duration, "Duration should match expected value=%f, actual=%f", file.Duration, format.DurSecs)
+		assert.True(t, format.FPS == file.FPS, "FPS should match expected value=%f, actual=%f", file.FPS, format.FPS)
```
j0sh, line 1911: The `files` table is nice because it means a sub-test would work well here.  Things can also be simplified a bit, eg we can remove the timing / logging (does not add anything) and I don't think that GetCodecInfoBytes really needs to be called - it simply overwrites the original GetCodecInfo results without checking them, and this PR doesn't directly change GetCodecInfoBytes anyway

```suggestion
		t.Run(file.Filename, func(t *testing.T) {
			assert := assert.New(t)
			status, format, err := GetCodecInfo(path.Join(dir, file.Filename))
			assert.Nil(err, "getcodecinfo error")
			assert.Equal(CodecStatusOk, status, "status not ok")
			assert.Equal(file.Duration, format.DurSecs, "duration mismatch")
			assert.Equal(file.FPS, format.FPS, "fps mismatch")
		}
```

ffmpeg/ffmpeg_test.go
```diff
@@ -1869,3 +1869,45 @@ func TestResolution_Clamp(t *testing.T) {
...
+
+func TestDurationFPS_GetCodecInfo(t *testing.T) {
+	run, dir := setupTest(t)
+	defer os.RemoveAll(dir)
+
+	//Generate test files
+	cmd := `
+	cp "$1/../data/bunny.mp4" test.mp4
+	cp "$1/../data/duplicate-audio-dts.ts" test.ts
+	ffmpeg -loglevel warning -i test.mp4 -vn -c:a flac test.flac
```
j0sh, line 1881: This is good, it would be nice to also add checks for expected values from ffprobe - helps ensure that our own calculations are correct

also just checking, which one of these triggers the no-duration case that requires manual copying? (that *should* also surface via ffprobe, eg a duration of "N/A")

ffmpeg/extras.c
eliteprox: r_frame_rate is working well, seems more reliable

ffmpeg/ffmpeg_test.go
eliteprox: It was with webm files. `ffprobe -show_streams test.webm | grep 'duration='` gives **N/A** for duration. I've added webm to the test, it is working with the latest changes.

ffmpeg/extras.c
```diff
@@ -163,6 +188,13 @@ int lpms_get_codec_info(char *fname, pcodec_info out)
   if(!audio_present && !video_present) {
     // instead of returning -1
     ret = GET_CODEC_STREAMS_MISSING;
+  } else if (video_present) {
+    out->dur = ic->duration / AV_TIME_BASE;
+  }
+
+  if (out->dur <= 0) {
+    //ic->duration / AV_TIME_BASE doesn't return accurate duration for audio-only streams, so we calculate the duration manually
+    out->dur = calculate_audio_dur(ic, astream);
```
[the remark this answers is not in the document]
eliteprox: The change above seems to have fixed the issue entirely, but there are still many instances where `calculate_stream_duration` must be used. I tried an mp3 file and it causes ic->duration to be `-9223372036854775808`. 

The change adding `ic->max_analyze_duration = 60 * AV_TIME_BASE;` did not have any effect, and I was able still process audio files that were hours long, so I added this change.

ffmpeg/extras.c
```diff
@@ -130,6 +130,31 @@ int lpms_rtmp2hls(char *listen, char *outf, char *ts_tmpl, char* seg_time, char
...
+  // Seek to the beginning of the audio stream
+  av_seek_frame(ic, astream, 0, AVSEEK_FLAG_BACKWARD);
+  while (av_read_frame(ic, &pkt) >= 0) {
+    if (pkt.stream_index == astream) {
+      if (pkt.pts != AV_NOPTS_VALUE) {
+        if (last_pts != AV_NOPTS_VALUE) {
+          // Calculate the difference between the current and last PTS
+          int64_t pts_diff = pkt.pts - last_pts;
+          // Convert the PTS difference to seconds and add to duration
+          duration += pts_diff * av_q2d(ic->streams[astream]->time_base);
```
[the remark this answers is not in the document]
eliteprox: Optimized `calculate_stream_duration` to calculate at the end and use the smallest and largest pts for first/last pts  https://github.com/eliteprox/lpms/blob/f2edd3788176030f0d77699bb0912f67b3e9d4d5/ffmpeg/extras.c#L135-L149

j0sh (changes requested):

ffmpeg/extras.c
j0sh: note that `r_frame_rate` does not interact well with VFR or otherwise unreliable streams (RTP / webrtc) - this caused a big issue in the past, see #393

the suggestion was more to use `r_frame_rate` if `avg_frame_rate` is unset, but I am not sure which one you want to trust more

ffmpeg keeps them separate for a reason - so the caller can distinguish between what's declared in the container, vis-a-vis a calculated approximation from the actual data

ffmpeg/extras.c
eliteprox: Removed `calculate_stream_duration`

ffmpeg/extras.c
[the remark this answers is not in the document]
eliteprox: Removed `calculate_stream_duration`

ffmpeg/ffmpeg_test.go
eliteprox: Added the ffprobe commands, shortened all the videos to 2 seconds. The int64 change is working well so I removed `calculate_stream_duration`. I am happy with rounding numbers for now

ffmpeg/extras.c
```diff
@@ -54,6 +54,7 @@ int lpms_rtmp2hls(char *listen, char *outf, char *ts_tmpl, char* seg_time, char
 
   ret = avformat_open_input(&ic, listen, NULL, NULL);
   if (ret < 0) r2h_err("segmenter: Unable to open input\n");
+  ic->max_analyze_duration = 60 * AV_TIME_BASE; // 60 seconds
```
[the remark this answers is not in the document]
eliteprox: removed

ffmpeg/ffmpeg_test.go
j0sh: >  I am happy with rounding numbers for now

Double checking (no pun intended) ... anything less than 1 (eg, 0.99) will be rounded down to 0, 1.99 becomes 1, etc. That seems pretty significant (eg, how do you distinguish between "very short" and "zero duration") but if is still OK for your purposes, then 👍 

Otherwise, I believe we can still match the values [calculated by ffprobe](https://github.com/FFmpeg/FFmpeg/blob/master/fftools/ffprobe.c#L952) - our equivalent would be `out->dur = ic->duration * av_q2d(AV_TIME_BASE_Q)` ... mathematically the same as the old code, but I don't know if the computed precision would differ (and it would only come into play for very large durations, anyhow)

ffmpeg/extras.c
```diff
@@ -164,6 +164,9 @@ int lpms_get_codec_info(char *fname, pcodec_info out)
     // instead of returning -1
     ret = GET_CODEC_STREAMS_MISSING;
   }
+  if (ic->duration != AV_NOPTS_VALUE) {  
+    out->dur = ic->duration / AV_TIME_BASE;
```
j0sh, line 168: FYI you might want to configure your editor to highlight trailing whitespace because there is some here

```suggestion
  if (ic->duration != AV_NOPTS_VALUE) {
    out->dur = ic->duration / AV_TIME_BASE;
```

ffmpeg/ffmpeg_test.go
```diff
@@ -1869,3 +1869,53 @@ func TestResolution_Clamp(t *testing.T) {
...
+	defer os.RemoveAll(dir)
+
+	//Generate test files
+	cmd := `
+	cp "$1/../data/duplicate-audio-dts.ts" test.ts
+	ffprobe -loglevel warning -show_format test.ts | grep duration=
+	ffprobe -loglevel warning -show_streams -select_streams v test.ts | grep r_frame_rate=
+
+	cp "$1/../data/bunny.mp4" test.mp4
+	ffmpeg -loglevel warning -i test.mp4 -c:v copy -t 2 -c:a copy -t 2.12564 test-short.mp4
```
j0sh, line 1884: `-t` repeated twice here, which one should be used?

ffmpeg/ffmpeg_test.go
```diff
@@ -1869,3 +1869,53 @@ func TestResolution_Clamp(t *testing.T) {
...
+	ffprobe -loglevel warning -show_format test-short.mp4 | grep duration=
+	ffprobe -loglevel warning -show_streams -select_streams v test-short.mp4 | grep r_frame_rate=
+
+	ffmpeg -loglevel warning -i test-short.mp4 -c:v libvpx -c:a vorbis -strict -2 test.webm
```
j0sh, line 1888: this is the one line that really could use `-t` ... copying is fast, encoding audio is fast, but encoding video is slow

```suggestion
	ffmpeg -loglevel warning -i test-short.mp4 -c:v libvpx -c:a vorbis -strict -2 -t 2 test.webm
```

ffmpeg/ffmpeg_test.go
```diff
@@ -1869,3 +1869,53 @@ func TestResolution_Clamp(t *testing.T) {
...
 }
+
+func TestDurationFPS_GetCodecInfo(t *testing.T) {
+	run, dir := setupTest(t)
+	defer os.RemoveAll(dir)
+
+	//Generate test files
+	cmd := `
+	cp "$1/../data/duplicate-audio-dts.ts" test.ts
+	ffprobe -loglevel warning -show_format test.ts | grep duration=
```
j0sh, line 1880: please match against the actual duration / frame rate values for these ffprobes as the other testcases do ...

to get the actual values, put an `exit 1` at the end of this script to induce a failure and run the test. the lines matched by grep will be printed, put those in here, then remove the `exit` and hopefully everything passes

ffmpeg/ffmpeg_test.go
eliteprox: Updated to have it only one

ffmpeg/ffmpeg_test.go
eliteprox: Added the expected ffprobe duration and fps values to the grep command

eliteprox: Is there a way to pass/fail the test based on the values from ffprobe or is this good? https://github.com/livepeer/lpms/pull/407/commits/c85f8176a93bc6e89904d50bc7bcfdc187138740

j0sh (approved):
LGTM! 🚢

please squash when merging

ffmpeg/ffmpeg_test.go
j0sh: if the grep does not match anything, it will fail (nonzero exit code) so this is good

rickstaa:
@eliteprox very clean pull request 🚀. @j0sh thanks for helping to review this 🙏🏻.

PR #414, opened 2024-07-31
ffmpeg: Estimate duration for some audio formats in GetCodecInfoBytes

j0sh:
not all formats have a duration baked into the container. in those cases, ffmpeg normally uses the file size and bitrate to estimate the duration, but pipes do not have a file size ... however, we do know the file size from the input parameter (a byte slice, `data`) so calculate the duration ourselves for a few audio formats.

Also add a couple of small test files to cover codecs where we do not compile in encoders (mp3 and opus), and extend the existing test cases to also call GetCodecInfoBytes

Note that certain formats are known not to work with this method, eg ogg does not return a bitrate at all. This is  covered in test cases as well.


eliteprox:
Tested these changes on the `audio-to-text` Whisper pipeline and it is working excellent. Great work!

rickstaa:
@j0sh I checked the code and everything looks correct. I could not test this pull request directly but https://github.com/livepeer/lpms/pull/415 which worked perfectly.

rickstaa (approved):
Code and tests look good. Tested it out using https://github.com/livepeer/lpms/pull/415 since I'm on Ubuntu 24 which doesn't easily allow running the lower ffmpeg version.

PR #415, opened 2024-07-31
Merge AI Video commits into master

j0sh:
The FPS / duration tests seem to have been written against a newer ffmpeg than what the LPMS AI Video branch is compiled with, which is causing them to fail locally, and for some reason CI is not running so we did not catch this earlier ... so let's port over the AI Video changes to master and work directly off master, to minimize further drift.

Cherry-picked the following from the [ai-video](https://github.com/livepeer/lpms/tree/ai-video) branch:

* [ffmpeg: Use helper to check for video metadata](https://github.com/livepeer/lpms/commit/481c24fca1e129802392395e837bfcff0f587b18)
* [ffmpeg: Support image2 demuxer](https://github.com/livepeer/lpms/commit/de94555cdc69ea566f9f8770c5e3f26af10a67ca)
* [Add media duration to lpms_get_codec_info for GetCodecInfo (](https://github.com/livepeer/lpms/commit/a17d3c6d4c522f582524ab1782c0fa673c8f9f31)https://github.com/livepeer/lpms/pull/407[)](https://github.com/livepeer/lpms/commit/a17d3c6d4c522f582524ab1782c0fa673c8f9f31)
* https://github.com/livepeer/lpms/pull/414

Also since LPMS owns `install_ffmpeg.sh` now there is an install_ffmpeg.sh update in as the first commit, using the [diff](https://github.com/livepeer/go-livepeer/pull/3078/files#diff-4ae778054809274731b9da0c6a5a869c0bd214e92f954a5c9c39181748c2f175) of the go-livepeer ai-video branch against master as a reference:
* [install_ffmpeg: Compile in more stuff for AI](https://github.com/livepeer/lpms/commit/1ff87e4e3cfb5e5c9d9c17659f06cd2b6e442d0e)

PR #418, opened 2024-08-16
Fix CUVID crash on resolution change

j0sh:
Also adds a bunch of other changes necessary to better support
mid-stream resolution changes.

Unfortunately with CUVID there still seems to be a brief flash of
green (looks to be the length of the decoder's internal frame buffer)
but we can tackle that separately. This PR simply makes the transcoder
1) not crash, and
2) correctly encode mid-stream rotations, including with CPUs

Depends on https://github.com/livepeer/lpms/pull/417

The following commits can all stand alone except the last one (unit tests) depends on all of them. I can make separate PRs if that is preferable for ease of review. Let's also keep these commits separate when merging.

[ffmpeg: Flush filters before re-initialization.](https://github.com/livepeer/lpms/commit/dd12875e9b5ce2114b6aa1be2ae1a35f764cb296) 
Also add another condition for re-initialization: if the
input resolution changes. This triggers the filter graph
to re-build and adjust to the new resolution, when CPU
encoders are in use. Resolves a long-standing TODO.

[ffmpeg: Re-init encoder on resolution change.](https://github.com/livepeer/lpms/commit/133d1832e977d87ea63d8351a5c1e34d51c4cfd6)
Necessary to have the encoder actually pick up the new
resolution correctly.

[ffmpeg: Reset the flush packet after each keyframe.](https://github.com/livepeer/lpms/commit/26fdf8cb1af355565b9e81bea6bc9c25b6fa0587) 
This handles cases where the packet may contain a frame
that triggers a decoder reset - we do not want to cause
a reset during the flushing process.

[ffmpeg: Handle EAGAIN from decoder and drain](https://github.com/livepeer/lpms/commit/6db145ea5a96d11dd1b27200f6990ab74f3f0d18) 
This usually happens with CUVID if the decoder needs to be reset
internally for whatever reason, such as a mid-stream resolution
change. Previously, EAGAIN would loop and cause a crash.

Also block demuxing until decoder is ready to receive packets again.

[ffmpeg: Add tests for rotation.](https://github.com/livepeer/lpms/commit/986324ce939c99298c736d59f277a084bfadb0c6)

### DIFF
--- a/ffmpeg/encoder.c
+++ b/ffmpeg/encoder.c
@@ -255,7 +255,11 @@ int open_output(struct output_ctx *octx, struct input_ctx *ictx)
 	if(strcmp(octx->xcoderParams,"")!=0){
 	    av_opt_set(vc->priv_data, "xcoder-params", octx->xcoderParams, 0);
 	}
-    ret = avcodec_open2(vc, codec, &octx->video->opts);
+    // copy codec options and open encoder
+    AVDictionary *opts = NULL;
+    if (octx->video->opts) av_dict_copy(&opts, octx->video->opts, 0);
+    ret = avcodec_open2(vc, codec, &opts);
+    if (opts) av_dict_free(&opts);
     if (ret < 0) LPMS_ERR(open_output_err, "Error opening video encoder");
     octx->hw_type = ictx->hw_type;
   }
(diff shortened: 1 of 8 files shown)


ffmpeg/encoder.c
```diff
@@ -332,12 +336,83 @@ int reopen_output(struct output_ctx *octx, struct input_ctx *ictx)
...
+        //fprintf(stderr, "JOSH - flushing encoder\n");
+        if (!ret) {
+          if (!octx->fps.den && octx->vf.active) {
+            // adjust timestamps for filter passthrough
+            time_base = octx->vf.time_base;
+            int64_t pts_dts = pkt->pts - pkt->dts;
+            pkt->pts = (int64_t)pkt->opaque; // already in filter timebase
+            pkt->dts = pkt->pts - av_rescale_q(pts_dts, encoder->time_base, time_base);
+          }
+          //fprintf(stderr, "JOSH - flushing %"PRId64"/%"PRId64"\n", pkt->dts, pkt->pts);
```
mjh1, line 396: probably want to remove these commented out lines

ffmpeg/ffmpeg_test.go
```diff
@@ -1927,3 +1927,248 @@ PTS_EOF
...
+			Oname:   dir + "/out-double-rotated.ts",
+			Accel:   accel,
+		}})
+	require.NoError(t, err)
+
+	assert.Equal(t, 360, res.Decoded.Frames)
+	assert.Equal(t, 181, res.Encoded[0].Frames) // should be 180 ... ts rounding ?
+	assert.Equal(t, 360, res.Encoded[1].Frames)
+
+	// TODO test rollover of gop interval during flush
```
mjh1, line 1996: I'm not sure what we normally do in this repo for TODOs but you maybe want to raise tickets instead if they're important to come back to so we don't forget.

ffmpeg/encoder.c
j0sh: good catch, thanks - fixed in 4175186fe5718b6c0ed0cc655c1856bfd914c707

ffmpeg/ffmpeg_test.go
j0sh: this specific note is not so important, it is more noting a gap in the test coverage

the other TODOs sprinkled around these tests relate to discrepancies between cpu/gpu behavior are likely related to the green screen flash which we already have ticket(s) for

thomshutt (approved):

mjh1 (approved):
Looks good but I'm afraid I don't really know this codebase and C sorry

j0sh:
Squashed the comments from the PR review, changed the merge target to `master` so we will also pull in #417, then rebased onto latest master (I had been working off an older tip which had merge conflicts)

The CI-driven unit tests will probably fail right now even though they pass locally; see https://github.com/livepeer/lpms/pull/415#issuecomment-2276392076

PR #421, opened 2024-09-05
ffmpeg: Add a metadata option to each output

j0sh:
This will allow us to identify the software version and who transcoded a given segment.


leszko (approved):

mjh1 (approved):

j0sh:
Missed adding the metadata when re-opening just the muxer. This affected only nvenc with multiple transcodes, so also added tests specifically around that. Fixed in https://github.com/livepeer/lpms/pull/421/commits/beda797c2cbd393eb65d81e503419835da06d5a4

j0sh:
Tests pass locally; CI has been broken since the AI video merge but that is being debugged separately https://github.com/livepeer/lpms/pull/422

PR #420, opened 2024-08-28
implement looping, demuxer options

iameli-streams:
So -stream_loop doesn't work when provided as a demuxer option, as it's implemented in https://github.com/FFmpeg/FFMpeg/blob/master/fftools/ffmpeg_demux.c and we handle our own demuxing. Thankfully, this seemed to be relatively easy to port over using the preexisting work that we have for handling discontinuity. Lightly tested.

Demuxer options: Pretty straightforward, follows the same syntax as the output muxer options.

My use case: I've already got a livestream being sliced up into a bunch of C2PA-signed MP4 files. But now I need to turn them back into an MKV stream. To do this I'm using a hack on top of ffmpeg's `concat` muxer, looks like this on the command line:
```
ffmpeg \
  -stream_loop -1 \
  -safe 0 \
  -protocol_whitelist file,http,https,tcp,tls \
  -i http://127.0.0.1:9090/playback/0xf081d6383777482868faa8d5534a5f1a7777bee8/concat \
  -map 0 \
  -c copy \
  -y output.mkv
```

And that concat file looks like this:
```
ffconcat version 1.0
file 'http://127.0.0.1:9090/playback/0xf081d6383777482868faa8d5534a5f1a7777bee8/latest.mp4'
file 'http://127.0.0.1:9090/playback/0xf081d6383777482868faa8d5534a5f1a7777bee8/latest.mp4'
```
And those `latest.mp4` files get 301 redirected to the correct MP4 files. Works great!

PR #426, opened 2024-11-18
ffmpeg: Add demuxer options.

j0sh:
This adds demuxer options as a complement to the existing [encoder/muxer](https://github.com/livepeer/lpms/blob/ffde2327537517b3345162e9544704571bc58a34/ffmpeg/ffmpeg.go#L112-L114) options which allows us to

1. explicitly select the demuxer to use if probing doesn't return a good result, and
2. configure the demuxer with additional options

This has come up a few times while looking at various things so it is good to have an API that is fully configurable out of the box.

This was partially implemented in #420 but this is more complete - it actually allows for demuxer selection and adds test cases.

### DIFF
--- a/ffmpeg/decoder.c
+++ b/ffmpeg/decoder.c
@@ -355,10 +355,19 @@ int open_input(input_params *params, struct input_ctx *ctx)
 
   ctx->transmuxing = params->transmuxing;
 
-  // open demuxer/ open demuxer
+  const AVInputFormat *fmt = NULL;
+  if (params->demuxer.name) {
+    fmt = av_find_input_format(params->demuxer.name);
+    if (!fmt) {
+      ret = AVERROR_DEMUXER_NOT_FOUND;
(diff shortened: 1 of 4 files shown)


leszko (approved):
@j0sh Could you add a description to the PR to add the context ( 1) Why this change? 2) What is it changing from the "user" perspective? ). Thanks!

j0sh:
@leszko 

Updated the PR description. For more color as to "why" : I have been testing some flags to tune for latency but the gains are marginal so I am not sure we will keep them. eg, `nobuffer` allows us to drop up a few frames at the beginning to stay closer to the live edge, but this only really helps if you're playing back from the very beginning. if you start playback mid-stream then you're already starting at the live edge and it doesn't do anything.

BTW tests seem to be failing so I will hold off on merging this until I have some time to make CI green.

j0sh:
Tests are passing locally; the CI failures are unrelated, see https://github.com/livepeer/lpms/pull/421#issuecomment-2338642277

PR #427, opened 2024-11-21
Update ffmpeg.go  change preset from slow to medium

Tibia2000:
Updated the FFmpeg preset from slow to medium because the increased computation time required for slow does not provide a significant visual quality improvement. This change optimizes the balance between performance and quality.
image


j0sh (approved):

j0sh:
LGTM but would also like input from @leszko since he is the one who made the original change to improve quality

Also wanted to confirm that this _does_ result in a measurable speedup for GPU with our benchmark tools?

Tibia2000:
> LGTM but would also like input from @leszko since he is the one who made the original change to improve quality
> 
> Also wanted to confirm that this _does_ result in a measurable speedup for GPU with our benchmark tools?

Yes it does, performance hit with slow is over 50%:
image
image

leszko (approved):
That's fine, we can change it to "medium"

PR #428, opened 2024-12-02
install_ffmpeg: Add FLV muxer

j0sh:
This should make RTMP output work on production (non-debug) builds of LPMS

### DIFF
--- a/install_ffmpeg.sh
+++ b/install_ffmpeg.sh
@@ -205,7 +205,7 @@ fi
 
 if [[ $BUILD_TAGS == *"debug-video"* ]]; then
   echo "video debug mode, building ffmpeg with tools, debug info and additional capabilities for running tests"
-  DEV_FFMPEG_FLAGS="--enable-muxer=md5,flv --enable-demuxer=hls --enable-filter=ssim,tinterlace --enable-encoder=wrapped_avframe,pcm_s16le "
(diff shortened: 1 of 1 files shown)


j0sh:
Tests are still broken in CI but pass locally

```
josh@josh-gpu:~/lpms$ time go test --tags=nvidia -timeout 1h ./ffmpeg
ok      github.com/livepeer/lpms/ffmpeg 1575.155s

real    26m16.094s
user    36m18.451s
sys     1m49.643s
```

install_ffmpeg.sh
```diff
@@ -205,7 +205,7 @@ fi
 
 if [[ $BUILD_TAGS == *"debug-video"* ]]; then
   echo "video debug mode, building ffmpeg with tools, debug info and additional capabilities for running tests"
-  DEV_FFMPEG_FLAGS="--enable-muxer=md5,flv --enable-demuxer=hls --enable-filter=ssim,tinterlace --enable-encoder=wrapped_avframe,pcm_s16le "
```
victorges, line 208: Why remove this one? Are those only extra from the non-dev ones?

victorges (approved):
LGTM!

install_ffmpeg.sh
j0sh: Yes, the base configuration includes flv now, so specifying it (again) with the dev flags isn't necessary (trying to keep things tidy)

PR #431, opened 2025-01-15
Use ffmpeg parser for H.264

j0sh:
Fixes a number of things with the nvidia built-in parser (cuvid) including a LPMS crash, choppy video quality, green screen flashing during rotation, inconsistent frame counts vs software decoding, etc. We also apparently gained GPU support for MPEG2 decoding.

This is a massive change: we can no longer add outputs up front due to the ffmpeg hwaccel API, so we have to wait until we receive a decoded video frame in order to add outputs. This also means properly queuing up audio and draining things in the same order.

This also unfortunately adds ~13 MB of new samples to the repo for test cases, since these are hard to generate synthetically.


ffmpeg/nvidia_test.go
```diff
@@ -23,15 +24,16 @@ func TestNvidia_BadCodecs(t *testing.T) {
 	run, dir := setupTest(t)
 	defer os.RemoveAll(dir)
 
-	fname := dir + "/mpeg2.ts"
-	oname := dir + "/out.ts"
+	fname := dir + "/test.flv"
+	//oname := dir + "/out.ts"
```
leszko, line 28: Nit: Remove commented-out code

leszko (approved):
I skimmed through the PR and looks good. Saying that, I don't have enough LPMS knowledge to review it in depth, so if you need a "better" review, then you can wait for approves from some other folks.

ffmpeg/nvidia_test.go
j0sh: Thanks for the spot, fixed and cleaned up in 9e51cee9fd2d1e6d14169784ed898a63d0caa960

ffmpeg/nvidia_test.go
```diff
@@ -603,6 +603,7 @@ func TestNvidia_API_AlternatingTimestamps(t *testing.T) {
 	tc := NewTranscoder()
 	idx := []int{1, 0, 3, 2}
 	for _, i := range idx {
+		// TODO this breaks with nvidia acceleration on the input!
```
j0sh, line 606: Note that this is _not_ a new problem with this PR; the old code had the same issue

mjh1:
> I skimmed through the PR and looks good. Saying that, I don't have enough LPMS knowledge to review it in depth, so if you need a "better" review, then you can wait for approves from some other folks.

Maybe @emranemran you could take a look at the C code? Can't remember if you've reviewed much of lpms in the past..

mjh1 (approved):
Similar to Rafal I'm afraid I don't know enough about this stuff, maybe we could do a knowledge transfer call or something?

PR #433, opened 2025-04-18
Replace videolan.org and nasm.us dependecies from `install_ffmpeg.sh`

hjpotter92:
We've been hit more than once because those domains were down or SSL error or something else.


j0sh (approved):
Not sure if CI is using this file directly, but if they are using go-livepeer's script then that would need to be updated too: https://github.com/livepeer/go-livepeer/blob/master/install_ffmpeg.sh

PR #434, opened 2025-05-27
add duration check for input files

ad-astra-video:
Fix to test input segments for durations longer than 5 minutes. 

Fixes bug that created ridiculously large output segments.

[long-ts.log](https://github.com/user-attachments/files/20470410/long-ts.log)
[wrong-ts.log](https://github.com/user-attachments/files/20470411/wrong-ts.log)

cc @rickstaa

### DIFF
--- a/ffmpeg/ffmpeg_test.go
+++ b/ffmpeg/ffmpeg_test.go
@@ -2424,3 +2424,60 @@ func TestTranscoder_PNGDemuxerOpts(t *testing.T) {
 	assert.Equal(t, 3, res.Decoded.Frames)
 	assert.Equal(t, 180, res.Encoded[0].Frames)
 }
+
+func TestTranscode_DurationLimit(t *testing.T) {
+	run, dir := setupTest(t)
+	defer os.RemoveAll(dir)
+	cmd := `
+		ffmpeg -f lavfi -i color=c=blue:s=1280x720 -r 1 -frames:v 301 -c:v libx264 test-dur-bad.ts
+		ffmpeg -f lavfi -i color=c=blue:s=1280x720 -r 1 -frames:v 300 -c:v libx264 test-dur-good.ts
+	`
+	run(cmd)
+
+	// Create a transcoder instance
+	transcoder := NewTranscoder()
+	defer transcoder.StopTranscoder()
+
+	// Set up transcode options
+	badInput := &TranscodeOptionsIn{
+		Fname: fmt.Sprintf("%v/test-dur-bad.ts", dir),
+		Accel: Software,
+	}
+
+	goodInput := &TranscodeOptionsIn{
+		Fname: fmt.Sprintf("%v/test-dur-good.ts", dir),
+		Accel: Software,
+	}
+
+	profiles := []VideoProfile{
+		{
+			Name:       "test_profile",
+			Resolution: "854x480",
+			Bitrate:    "1000k",
+		},
+	}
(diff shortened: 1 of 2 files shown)


ad-astra-video:
@j0sh I added two tests. One tests good and bad segments using file inputs.  Second test uses a pipe input with 301 second input that does not trigger an error.

ffmpeg/ffmpeg_test.go
```diff
@@ -2424,3 +2425,111 @@ func TestTranscoder_PNGDemuxerOpts(t *testing.T) {
...
+		},
+	}
+
+	// transcode bad input
+	_, errBadInput := transcoder.Transcode(badInput, options)
+
+	// Check that the correct error was returned
+	if errBadInput != ErrTranscoderDuration {
+		t.Errorf("Expected ErrTranscoderDuration for video longer than 5 minutes, got %v", errBadInput)
+	}
```
j0sh, line 2475:
```suggestion
	assert.Equal(t, ErrTranscoderDuration, errBadInput)
```

ffmpeg/ffmpeg_test.go
```diff
@@ -2424,3 +2425,111 @@ func TestTranscoder_PNGDemuxerOpts(t *testing.T) {
...
+	if errGoodInput != nil {
+		t.Error(errGoodInput)
+	}
+}
+
+func TestTranscoder_NoDurationLimitBytes(t *testing.T) {
+	run, dir := setupTest(t)
+	defer os.RemoveAll(dir)
+
+	cmd := `ffmpeg -f lavfi -i color=c=blue:s=1280x720 -r 1 -frames:v 301 -c:v libx264 test-dur-bad.ts`
```
j0sh, line 2490: Same note as above about building the input via timestamp rewriting, also maybe name it "test-dur-long.ts" (since its a valid input in this case, not a bad one) and add a comment indicating why this is expected to pass, but file-based isn't.

ffmpeg/ffmpeg_test.go
```diff
@@ -2424,3 +2425,111 @@ func TestTranscoder_PNGDemuxerOpts(t *testing.T) {
 	assert.Equal(t, 3, res.Decoded.Frames)
 	assert.Equal(t, 180, res.Encoded[0].Frames)
 }
+
+func TestTranscode_DurationLimit(t *testing.T) {
+	run, dir := setupTest(t)
+	defer os.RemoveAll(dir)
+	cmd := `
+		ffmpeg -f lavfi -i color=c=blue:s=1280x720 -r 1 -frames:v 301 -c:v libx264 test-dur-bad.ts
+		ffmpeg -f lavfi -i color=c=blue:s=1280x720 -r 1 -frames:v 300 -c:v libx264 test-dur-good.ts
```
j0sh, line 2434: It is faster to use an existing sample and just modify it as needed, rather than trying to re-generate / re-encode new ones. Look at how the other test cases do it.

Eg:

```suggestion
		# generate a 1fps sample
		ffmpeg -i $1../transcoder/test.ts -c copy -bsf:v setts=ts=N/TB_OUT/1 -frames:v 301 -y test.ts
		# double check the sample actually has the characteristics we expect
		ffprobe -show_format test.ts  | grep duration=301.00
```

Also feels like we can probably skip the "good" tests? If we still want to exercise the boundary conditions, then encode to the smallest possible output file.

PR #436, opened 2025-08-19
Fix infinite output segments

stronk-dev:
This PR:

- Checks for huge or negative timestamp jumps in the filter and falls back to the default frame duration in such cases. This prevents the encoder from getting stuck in a loop processing specific inputs. 
Note: technically the check for negative timestamps should be enough, but included the other one as well. We could also omit it.
- Test that confirms whether one of the problematic input segments no longer triggers infinite outputs.


stronk-dev:
Note: Navigare reports he has used this build and it resulted in memory leaks. I don't see an apparent cause for this, but a point to keep in mind for the review. Might be missing something (like maybe it is something stupid like how I probe the input size)

stronk-dev:
Closed, in favour of #438 which targets the root cause

PR #437, opened 2025-09-03
Abort transcoding if output size exceeds 30x input size

stronk-dev:
This PR:

- Aborts processing if the output file exceeds 30x input file size.
- Test that confirms whether one of the problematic input segments no longer triggers infinite outputs.

Note that the test probably no longer works. Will fix.


j0sh:
Would it be cleaner to count output frames instead of output bytes?

stronk-dev:
> Would it be cleaner to count output frames instead of output bytes?

agreed.

I also found some other potential area's we could use to detect 'error states' (based off of logs from the error state from the test segment):
- the [amount of decoded frames](https://github.com/livepeer/lpms/blob/a53e20a3346616f56e2b1e9b97db9780ebbe029c/ffmpeg/transcoder.c#L426) remains static, while [encoded frames](https://github.com/livepeer/lpms/blob/a53e20a3346616f56e2b1e9b97db9780ebbe029c/ffmpeg/encoder.c#L423) kept increasing. Now I can imagine that timejumps just happen from time to time where some degree of duplication is desired, but we can think of some rules here to trigger an exit.
- The [original input PTS](https://github.com/livepeer/lpms/blob/a53e20a3346616f56e2b1e9b97db9780ebbe029c/ffmpeg/filter.c#L337) remained constant. So a solution here could be to track the previous source PTS and count 'duplicate frames' in the filter. Maybe error if there's a second worth of duplicate frames?

PR #438, opened 2025-09-03
Fix pts overflow in fps filter

j0sh:
This was causing some very large segments to be produced if the input had some weird characteristics like missing timestamps.

Alternative to https://github.com/livepeer/lpms/pull/436


j0sh:
@ad-astra-video 

image

thomshutt (approved):

